PR #7, opened 2021-05-28
Add contract addresses for Fantom testnet

fmrsabino:
Adds contract addresses for Fantom testnet (`chain_id = 4002`)


rmeissner:
Chain id can be found here https://docs.fantom.foundation/tutorials/set-up-metamask-testnet

PR #14, opened 2021-07-28
Add Arbitrum testnet deployments

fmrsabino:
It looks like the contracts were already deployed to the Arbitrum testnet. I ran `getCode` on every address using the Testnet RPC and every address returned non-empty code. I did not do a diff of the code to the other networks though.


rmeissner:
We did bot add arbitrum test networks as they were changing to often, even resulting in complete resets

fmrsabino:
> We did bot add arbitrum test networks as they were changing to often, even resulting in complete resets

I see. Will close this Pull Request then 👍

PR #18, opened 2021-09-22
Celo deployment

rmeissner:
Add addresses for contracts deployed to Celo. As Celo requires replay protection the contracts were deployed using https://github.com/gnosis/safe-contracts/pull/344.

To check deployment checkout `safe-contracts` on branch `feature/custom_singleton_factory` and set `CUSTOM_DETERMINISTIC_DEPLOYMENT` to `true`


Uxio0:
Shouldn't this contracts be verified on either blockscout (https://explorer.celo.org/) or at least sourcify (e.g. https://repo.sourcify.dev/contracts/full_match/42220/0x017062a1dE2FE6b99BE3d9d37841FeD19F573804/ )

rmeissner:
You can use the safe-contracts repo to verify the contracts. The verification on sourcify would be nice to have, but we should add better support for doing this.

Uxio0 (approved):
I've checked all the contracts manually, looks good

PR #33, opened 2021-11-15
Added Fantom network deployment addresses

HonzaDajc:


rmeissner (changes requested):
The addresses for the 1.1.1 version do not correspond to addresses that we expect. Please handle them in a separate PR and take a look at https://github.com/gnosis/safe-contract-deployment-replay

HonzaDajc:
> The addresses for the 1.1.1 version do not correspond to addresses that we expect. Please handle them in a separate PR and take a look at https://github.com/gnosis/safe-contract-deployment-replay

The addresses are different, because we deployed contracts manually, but using them for Safe. On these adresses there are Safes created and that's why I wanted to reference them here in the safe-deployments.

rmeissner:
Closed due to inactivity. Reopen if requested changes have been applied.

PR #40, opened 2021-11-30
Add IoTex testnet network

datradito:
Endpoints for verification:

4690 (IoTex Testnet): `https://babel-api.testnet.iotex.io/`


src/assets/v1.3.0/compatibility_fallback_handler.json
```diff
@@ -17,6 +17,7 @@
     "1285": "0xf48f2B2d2a534e402487b3ee7C18c33Aec0Fe5e4",
     "1287": "0xf48f2B2d2a534e402487b3ee7C18c33Aec0Fe5e4",
     "4002": "0xf48f2B2d2a534e402487b3ee7C18c33Aec0Fe5e4",
+    "4690": "0xc6Dd460C2a3f4cf74C2D213B4CB81c6943DA7793",
```
rmeissner, line 20: Which commit from the contracts repo did you use for deployment? The expected address should be either   `0xf48f2B2d2a534e402487b3ee7C18c33Aec0Fe5e4` or `0x017062a1dE2FE6b99BE3d9d37841FeD19F573804`

rmeissner (changes requested):

rmeissner:
Please make sure you use the right commit for the deployment. The addresses should be the same as some of the existing  networks. 

You should use the `1.3.0-libs.0` tag/release on the safe contracts repo for deployment.

src/assets/v1.3.0/compatibility_fallback_handler.json
datradito: Hello @rmeissner I used [https://github.com/gnosis/safe-contracts/tree/1.3.0-libs.0](https://github.com/gnosis/safe-contracts/tree/1.3.0-libs.0), but as @mikheevm mentioned here:
[https://discord.com/channels/477106835862716416/914975477389721611/914979975105495100](https://discord.com/channels/477106835862716416/914975477389721611/914979975105495100)

I had to set the `deterministicDeployment` property to false

image

on all the deployment scripts:
image

because the chain doesn't accept deterministic factory. Please @mikheevm correct me if I'm wrong about that.

rmeissner:
Could you please link a documentation that outlines why a contract factory is not supported. For networks that enforce eip-155 we use https://github.com/gnosis/safe-singleton-factory. Do get support for this new network please open an issue there (same as was done for Boba network for example).

We do not support deployments that are based on private keys that are not controlled by us.

datradito:
> 

Hi @rmeissner , perfect, I already opened an issue to add Iotex support:
[https://github.com/gnosis/safe-singleton-factory/issues/10](https://github.com/gnosis/safe-singleton-factory/issues/10)

datradito:
waiting to solve issue [https://github.com/gnosis/safe-singleton-factory/issues/10)](https://github.com/gnosis/safe-singleton-factory/issues/10)

PR #41, opened 2021-12-01
deterministc deployment of boba mainnet and rinkeby

InoMurko:


src/assets/v1.3.0/sign_message_lib.json
```diff
@@ -7,13 +7,15 @@
     "1": "0xA65387F16B013cf2Af4605Ad8aA5ec25a2cbA3a2",
     "4": "0xA65387F16B013cf2Af4605Ad8aA5ec25a2cbA3a2",
     "10": "0x98FFBBF51bb33A056B08ddf711f289936AafF717",
+    "28": "0xE1209DABA608dc6c07d018042a8B8Cdb784bE05b",
```
rmeissner, line 10: This address is different than expected. Is there a specific reason why?

rmeissner: We tried to verify the deployment by running `yarn deploy-custom` in the safe-contracts@v1.3.0-libs.0 and got a different address

InoMurko: uh uh,
could it be that 1.3.0 tag does not actually deploy sign_message_lib:
https://github.com/gnosis/safe-contracts/blob/v1.3.0/src/deploy/deploy_libraries.ts

I've added 
```
  await deploy("SignMessageLib", {
    from: deployer,
    args: [],
    log: true,
    deterministicDeployment: true,
  });
```
to deploy_libraries.ts.

InoMurko: https://github.com/omgnetwork/safe-contracts/blob/492ce791d5b93fb849e91d6468b590c8e09d6d99/src/deploy/deploy_libraries.ts

rmeissner: The latest release is [v1.3.0-libs.0](https://github.com/gnosis/safe-contracts/releases/tag/v1.3.0-libs.0) which includes this. I will make this more clear in the readme of the safe contracts to avoid confusion in the future.

rmeissner: Note: https://github.com/gnosis/safe-contracts/issues/361

InoMurko: thanks @rmeissner. 
looking at the commit diff, I don't see any meaningful contract changes.
if you verify the addresses towards 1.3.0-libs.0, do you get what you're supposed to get?

rmeissner: The problem is that some metadata generated by the solidity compiler changes (as the file was moved to a different). We would like to keep the addresses in sync, to avoid confusion for the users. So if possible it would be nice to deploy this contract as the expected address.

InoMurko: I think it's fixed! thank you for your patience @rmeissner !

fmrsabino (approved):

PR #45, opened 2021-12-10
changes to try deployed contract

arater:


rmeissner:
We do not have an official 1.4.0 version therefore this will not be added

PR #49, opened 2021-12-26
added support for local ganache-cli chain

vigan-abd:


rmeissner:
We currently do not plan to add the local ganache chain (or any other) to this repo, as this is meant only for non-dev networks.

vigan-abd:
alright, closing PR

PR #65, opened 2022-02-22
Closes #64: Fix repo related links

### LINKED ISSUE #64: Incorrect URLs in `package.json`
Some URLs in `package.json` point to the non-existing URL https://github.com/gnosis/gnosis-deployments, maybe because the repo was renamed after they were added. They should be (as of this writing) https://github.com/gnosis/safe-deployments instead.

In particular, fixing this would be useful because this appears to be the URL used by Dependabot to retrieve the changelog ([example](https://github.com/gnosis/cow-token/pull/124)).

rmeissner:
Closes #64

- Update `repository` property in `package.json`
- Update `bugs` property in `package.json`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gnosis.pm/safe-deployments",
-  "version": "1.9.0",
+  "version": "1.9.1",
   "description": "Collection of Safe singleton deployments",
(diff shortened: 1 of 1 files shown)


Uxio0 (approved):
Looks good

PR #71, opened 2022-03-25
Add velas networks

rmeissner:
Add Velas networks:

- Mainnet - 106 - `https://evmexplorer.velas.com/rpc` 
- Testnet - 111 - `https://evmexplorer.testnet.velas.com/rpc` 

These nodes might be slow. If you experience timeouts with hardhat see https://hardhat.org/config/


Uxio0 (approved):
I've just checked using python, all the contracts are deployed

PR #75, opened 2022-04-07
change npm license to MIT

mikhailxyz:
The license was inconsistent on github and npm. This PR aligns them to MIT license

**Question**
Should it be a major version bump?


dasanra (approved):

rmeissner:
I would bump the version separately. As we also want to add other things for a new version. Also acutally this is just correcting the licence not updating it, as the licence agreement in the package/ repo is the one that should be looked at.

PR #89, opened 2022-05-20
fix: compare deployment versions semantically

iamacook:
## Overview

This compares deployment versions with `semver` instead of by strict comparison, therefore allowing for metadata, i.e.

```ts
'1.3.0' === '1.3.0+L2' // false
semverSatisfies('1.3.0', '1.3.0+L2') // true
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
   },
   "devDependencies": {
     "@types/node": "^14.14.21",
+    "@types/semver": "^7.3.9",
     "@typescript-eslint/eslint-plugin": "^4.7.0",
     "@typescript-eslint/parser": "^4.7.0",
     "eslint": "^7.13.0",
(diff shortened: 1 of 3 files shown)


src/factories.ts
```diff
@@ -1,15 +1,13 @@
...
+import { findDeployment } from './utils'
 
 // This is a sorted array (newest to oldest)
 const factoryDeployments: SingletonDeployment[] = [
     ProxyFactory130, ProxyFactory111
 ]
 
 export const getProxyFactoryDeployment = (filter?: DeploymentFilter): SingletonDeployment | undefined => {
-    return findDeployment(applyFilterDefaults(filter), factoryDeployments)
+    return findDeployment(filter, factoryDeployments)
```
iamacook, line 12: A default filter parameter in `findDeployment` is now used instead of `applyFilterDefaults`. It seemed like a worthwhile refactor, but I am happy to revert it.

mikhailxyz (approved):

PR #90, opened 2022-05-27
Add smartBCH and smartBCH Amber testnet

mainnet-pat:
Based on deterministic deployment using safe singleton factory


moisses89 (approved):
Looks good

PR #111, opened 2022-08-25
Update proxy_factory.json

Atawalpa:
For the one who sent bnb on the wrong network it will be nice to have the possibility to recover the funds creating the same address on the BSC network


Atawalpa:
> The 1.1.1 contracts are not all available on BSC, therefore they will not be added to this repo. It is not possible to recover funds sent to a Safe that used these contracts.

Thank you so there's no way to recover or deploy again BNB sent on this contract on version 1.1.1 even if we re create the entire proxy factory. when you said not all available on BSC witch one is available on BSC? THanks again!

rmeissner:
The Safe singleton (the logic contract) for 1.1.1 is not available on BSC therefore even if you recreate a Safe created via `createProxyWithNonce` you will not have access to it. It is possible to recreate Safes via `createProxy` (as they don't use the create2). But as this is only limited availability of the Safe contracts we will not add them to this repo.

Atawalpa:
So you can't help me to recover those funds correct?

PR #113, opened 2022-09-19
add ethereumpow and etherfair chain

benjamintshi:
Network Name: Ethereum Pow
New RPC URL: https://mainnet.ethereumpow.org
Chain ID: 10001
Currency Symbol (Optional): ETHW
Block Explorer URL: https://www.oklink.com/en/ethw/

Network Name: Ethereum Fair
New RPC URL: https://rpc.etherfair.org
Chain ID: 513100
Currency Symbol (Optional): ETF
Block Explorer URL: https://explorer.etherfair.org/


moisses89:
Currently Ethereum Pow is not listed in chainlinst and `10001` is the chainid for other network https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-10001.json

moisses89:
Ehtereum Fair seems to be ok, could you open a new PR with Ethereum Fair addresses and only for v1.3.0?

PR #115, opened 2022-09-20
add BOBA Avalanche L2 network (43288)

dirtycajunrice:


moisses89:
The provided chainId doesn't exist in https://github.com/ethereum-lists/chains/tree/master/_data/chains , we are not accepting PR if the chain is not included. 
I would like to comment also the following: 
- this should be divided in two different PR's one for v1.1.1 and other for v1.3.0.
- if you want to add the `v1.1.1` contract addresses you should provide all addresses not only one.   

But before to do anything the chain should be listed.

moisses89:
Please, also fill the following issue https://github.com/safe-global/safe-deployments/issues/new?assignees=&labels=%5Benhancement%5D&template=new_chain.yaml&title=%5BNew+chain%5D%3A+ and link with the PR when ready for review.

rmeissner (changes requested):
The addresses proposed differ from the expected addresses. This is a hard blocker. Please make sure you deploy the correct contracts

InoMurko:
Was merged https://github.com/safe-global/safe-deployments/commit/ee7640aff5f787d44e797c49cad6a52c5750ea99

PR #128, opened 2022-10-13
Add coston and songbird flare networks

pkakelas:
In both networks pre-eip-155 transactions were not supported thus I submitted https://github.com/safe-global/safe-singleton-factory/issues/65 and https://github.com/safe-global/safe-singleton-factory/issues/66 about a month ago.

After a couple of weeks and since the contracts were not deployed in those networks, I proceeded with a non-deterministic deployment and verified all the contracts.

Flare Songbird: https://flare.xyz/introducing-songbird/
Flare Coston: https://flare.xyz/coston-test-network/

Thanks folks!


moisses89:
@pkakelas we are not accepting non deterministic address, anyway we are working on your issue to add the singleton-factories.

pkakelas:
Thank you! Appreciate it 🥇

Issue #150, opened 2022-11-09
Add existing 1.0.0 contracts


mikhailxyz:
### ChainId

1

### Chainlist

1

### Contract deployed

- [X] I confirm that the contracts are deployed
- [x] [Add ProxyFactory v1.0.0](https://github.com/safe-global/safe-deployments/pull/151)
- [x] [Add MultiSend v1.0.0](https://github.com/safe-global/safe-deployments/pull/152)

### Version of contracts deployed

v1.0.0

### Relevant information

Currently, we only have the GnosisSafe contract address for 1.0.0, and it blocks some projects that are looking to support a wide range of safe versions and rely on this repo to get contract addresses. We should add all 1.0.0 contracts

germartinez:
@mikhailxyz
Not only for `chainId` = 1.
Also 4, 5, 42 and 100

PR #152, opened 2022-11-09
Add MultiSend v1.0.0

germartinez:
This PR adds the MultiSend v1.0.0 contract.

The Safe Core SDK will add support for Safe contracts v1.0.0 and this asset was missing.

Check: https://github.com/safe-global/safe-deployments/issues/150

### DIFF
--- a/src/assets/v1.0.0/multi_send.json
+++ b/src/assets/v1.0.0/multi_send.json
@@ -1,0 +1,29 @@
+{
+  "defaultAddress": "0xD4B7B161E4779629C2717385114Bf78D612aEa72",
+  "released": true,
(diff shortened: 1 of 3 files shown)


mikhailxyz (approved):

src/assets/v1.0.0/multi_send.json
```diff
@@ -0,0 +1,29 @@
+{
+  "defaultAddress": "0xD4B7B161E4779629C2717385114Bf78D612aEa72",
```
rmeissner, line 2: I would not add this multisend. It has a bug and can be selfdestructed.

PR #170, opened 2022-12-13
added haqq mainnet and testnet sc deployments

Hirama:
Request for adding #171 #172 
Mainnet - https://chainlist.org/chain/11235
Testnet - https://chainlist.org/chain/54211


mikhailxyz:
Why do the addresses differ on mainnet and testnet? They should be identical because the deployment is deterministic

Hirama:
> Why do the addresses differ on mainnet and testnet? They should be identical because the deployment is deterministic

Because on testnet is enabled reply protection. https://github.com/safe-global/safe-contracts#replay-protection-eip-155

mikhailxyz:
> > Why do the addresses differ on mainnet and testnet? They should be identical because the deployment is deterministic
> 
> Because on testnet is enabled reply protection. https://github.com/safe-global/safe-contracts#replay-protection-eip-155

have you seen [safe-singleton-factory](https://github.com/safe-global/safe-singleton-factory)? it's mentioned in the docs section you shared

Hirama:
> > > Why do the addresses differ on mainnet and testnet? They should be identical because the deployment is deterministic
> > 
> > 
> > Because on testnet is enabled reply protection. https://github.com/safe-global/safe-contracts#replay-protection-eip-155
> 
> have you seen [safe-singleton-factory](https://github.com/safe-global/safe-singleton-factory)? it's mentioned in the docs section you shared

Yes, so we need to provide signed tx to safe-singleton-factory repo and after that step we should prepare pr to safe-deployments. Right?

mikhailxyz:
> > > > Why do the addresses differ on mainnet and testnet? They should be identical because the deployment is deterministic
> > > 
> > > 
> > > Because on testnet is enabled reply protection. https://github.com/safe-global/safe-contracts#replay-protection-eip-155
> > 
> > 
> > have you seen [safe-singleton-factory](https://github.com/safe-global/safe-singleton-factory)? it's mentioned in the docs section you shared
> 
> Yes, so we need to provide signed tx to safe-singleton-factory repo and after that step we should prepare pr to safe-deployments. Right?

You just need to create an issue in the safe-singleton-factory for the safe team to sign a transaction to deploy the factory to testnet

Hirama:
@mikhailxyz thank you! I will prepare new pr.

PR #220, opened 2023-04-06
Add Tbc chain

frankie060392:


Uxio0:
There's no TBC chain on 1337, that chain id is used for testing

PR #228, opened 2023-04-23
Create a Safe contracts deployments summary table

ademidun:
To make it easy to see the deployed Safe contract addresses with their human-readable names and networks in one place. Includes links to the contracts on block explorers.

`chains.json` was taken from https://github.com/ethereum-lists/chains

safe-deployments-2


package.json
```diff
@@ -51,6 +51,7 @@
     "typescript": "^4.2.4"
   },
   "dependencies": {
+    "fs": "^0.0.1-security",
```
mmv08, line 54: "fs" is a part of node's stdlib, so this is redundant

src/generateSummary.ts
```diff
@@ -0,0 +1,53 @@
+import * as fs from 'fs';
```
mmv08, line 1: This file doesn't have to be a part of the source code and published to npm because it's only used to generate a summary file on GitHub

chains_sample.json
```diff
@@ -0,0 +1,329 @@
+[
```
mmv08, line 1: this sample file doesn't seem to be used anywhere, do we need it?

README.md
```diff
@@ -4,7 +4,9 @@
 
 This contract contains a collection of deployments of the contract of the [Safe contracts repository](https://github.com/safe-global/safe-contracts). 
 
-For each deployment the address on the different networks and the abi files are available. To get an overview of the available versions check the available [json assets](./src/assets/).
+For each deployment the address on the different networks and the abi files are available. To get an overview of the available versions check the available [json assets](./src/assets/) and [Summary Table](./SUMMARY.md).
+
+To regenerate the summary table to include new deployments run: `ts-node src/generateSummary.ts`.
```
mmv08, line 9:
```suggestion
To regenerate the summary table to include new deployments run: `npx ts-node src/generateSummary.ts`.
```

This can be run with npx and removing the need for global installation of `ts-node`

src/generateSummary.ts
```diff
@@ -0,0 +1,53 @@
...
+const description = 'This table contains a list of deployed Safe contracts. Chain information was taken from [Ethereum Lists: Chains](https://github.com/ethereum-lists/chains).\n';
+const versions = ['1.3.0', '1.2.0', '1.1.1', '1.0.0'];
+
+// filter chains by if their Chain ID is in the list of deployed chains
+const chains = ALL_CHAINS.filter((chain) => getProxyFactoryDeployment({ network: chain.chainId.toString() }));
+
+const headerRow = `| **Chain**                   | ${versions.map((version) => `**${version}**`.padEnd(27, ' ')).join('|')} |`;
+const seperatorRow = `| ${'-'.repeat(28)} | ${versions.map(() => ` ${'-'.repeat(26)} |`).join('')} `;
+
+const deployments = [getProxyFactoryDeployment, getSafeSingletonDeployment, getSafeL2SingletonDeployment];
```
mmv08, line 17: shall this include other deployments such as libraries and fallback handlers as well?

src/generateSummary.ts
```diff
@@ -0,0 +1,53 @@
...
+
+const chainRows = chains
+  .filter((chain) => {
```
mmv08, line 20: Are the chains already filtered at line 12?

mmv08:
I like the idea - the script just needs to be a little polished

src/generateSummary.ts
ademidun: I agree that it doesn't need to be published to npm but it should be part of the source code so others can generate the summary table if they add new chains.

chains_sample.json
ademidun: Agreed. I've removed it. I originally had it there in case people want to test the script without running on the entire JSON file since it's quite large. But people can just create it on their own if they need it.

package.json
ademidun: That's what I thought but it wasn't working for me until I did `yarn add fs` but I will remove it.

ademidun: Removed: https://github.com/safe-global/safe-deployments/pull/228/commits/022ea896e8732cd3a0de55abec6a37b5d1741449

README.md
ademidun: Tried `npx` but it wasn't generating the table. I've added it as an option.

src/generateSummary.ts
ademidun: Fixed

src/generateSummary.ts
ademidun: I wanted to share this with the team first before adding other deployments. I've added it now.

README.md
```diff
@@ -4,7 +4,9 @@
 
 This contract contains a collection of deployments of the contract of the [Safe contracts repository](https://github.com/safe-global/safe-contracts). 
 
-For each deployment the address on the different networks and the abi files are available. To get an overview of the available versions check the available [json assets](./src/assets/).
+For each deployment the address on the different networks and the abi files are available. To get an overview of the available versions check the available [json assets](./src/assets/) and [Summary Table](./SUMMARY.md).
+
+To regenerate the summary table to include new deployments run: `ts-node src/generateSummary.ts` or `npx src/generateSummary.ts`.
```
rmeissner, line 9: please add a script in the package.json for this.

src/generateSummary.ts
```diff
@@ -0,0 +1,54 @@
...
+
+import { getSafeL2SingletonDeployment, getSafeSingletonDeployment } from './safes';
+import { getProxyFactoryDeployment } from './factories';
+import { getDefaultCallbackHandlerDeployment, getCompatibilityFallbackHandlerDeployment, getFallbackHandlerDeployment } from './handler';
+import { getMultiSendDeployment, getMultiSendCallOnlyDeployment, getCreateCallDeployment, getSignMessageLibDeployment } from './libs';
+import ALL_CHAINS from '../chains.json';
+
+const title = '# Safe Deployments\n';
+const description = 'This table contains a list of deployed Safe contracts. Chain information was taken from [Ethereum Lists: Chains](https://github.com/ethereum-lists/chains).\n';
+const versions = ['1.3.0', '1.2.0', '1.1.1', '1.0.0'];
```
rmeissner, line 11: Maybe it would be better just to load the folder in the `assets` directory. This way the script doesn't need to be adjusted when a new version is added.

src/generateSummary.ts
rmeissner: What @mmv08 means is that it should not be in the `src` folder.

src/generateSummary.ts
```diff
@@ -0,0 +1,54 @@
...
+
+const headerRow = `| **Chain**                   | ${versions.map((version) => `**${version}**`.padEnd(27, ' ')).join('|')} |`;
+const seperatorRow = `| ${'-'.repeat(28)} | ${versions.map(() => ` ${'-'.repeat(26)} |`).join('')} `;
+
+const deployments = [getProxyFactoryDeployment, getSafeSingletonDeployment, getSafeL2SingletonDeployment,
+     getDefaultCallbackHandlerDeployment, getCompatibilityFallbackHandlerDeployment, getFallbackHandlerDeployment,
+     getMultiSendDeployment, getMultiSendCallOnlyDeployment, getCreateCallDeployment, getSignMessageLibDeployment];
+
+const chainRows = ALL_CHAINS
```
rmeissner, line 20: The current process is (if I understand correctly):
- Predefine all chains
- Predefine all versions
- Predefine all deployments
- Iterate nested over each to collect data
- Render md file.


This is quite a lot of predefined/hardcoded logic, which imo is not necessary.

It should be possible to do the following:
- load subfolders of `assets` -> versions
- load files of subfolders -> deployments
- Collect information into map of chain id to deployment
- Load chain information async
- Render md file

This way it would automatically work when new information is added (like a new version or new chain)

rmeissner (changes requested):
Thanks for the implementation. Couple comments:
- Could we remove the chains.json and load this information async from [Ethereum lists](https://github.com/ethereum-lists/chains)
- Please move `src/generateSummary.ts` to a new `scripts` folder as this file does not have to be uploaded to npm.
- It would make sense to run the script as part of pre publish and add the SUMMARY.md to npm (if this is not automatically the case).

README.md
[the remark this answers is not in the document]
ademidun: Not sure I understand what you mean?

ademidun: Done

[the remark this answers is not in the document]
mmv08: `npx src/generateSummary.ts` should be `npx ts-node src/generateSummary.ts`

ademidun:
> Could we remove the chains.json and load this information async from [Ethereum lists](https://github.com/ethereum-lists/chains)

I thought about this but chose not to do this because:
- Adds an external dependency
- Network Requests increase latency and slows down the script
- Can be added in a future PR as it's a non-blocking request outside the scope of what I intend this PR to be

> load files of subfolders -> deployments

I also thought about this but chose not to do this because:
- Requires complex parsing of JSON files and refactoring the entire code
- The current method utilizes the `safe-deployments` scripts and functions so it's more dynamic 
- Using the `safe-deployments` is also a way of testing `safe-deployments`
- The contract versions change infrequently enough that it's not that much effort to add a new function if we have a new contract deployment we want to add

In general, I think it's better to use functions like `getSafeL2SingletonDeployment` over interacting with raw JSON files.

README.md
ademidun: https://github.com/safe-global/safe-deployments/pull/228/commits/65be0d482d35a984599c4998492df2e5a1ce4204

src/generateSummary.ts
ademidun: Done: https://github.com/safe-global/safe-deployments/pull/228/commits/c2ad2b8335e576c9e010968627af6736aedee6d8

src/generateSummary.ts
ademidun: Done: https://github.com/safe-global/safe-deployments/pull/228/commits/65be0d482d35a984599c4998492df2e5a1ce4204

package.json
```diff
@@ -14,6 +14,8 @@
   ],
   "scripts": {
     "build": "yarn rimraf dist && tsc",
+    "summarize": "ts-node scripts/generateSummary.ts",
+    "summarize-npx": "npx ts-node scripts/generateSummary.ts",
```
rmeissner, line 18: This should not be required as ts-node is a dev dependency

mmv08: Actually, it's interesting that it's a dev dependency. It's not used anywhere

ademidun: Removed: https://github.com/safe-global/safe-deployments/pull/228/commits/1eecd7bda874c4f7a1e1ad41ecfec37ad1dbd210

src/generateSummary.ts
ademidun: Done: https://github.com/safe-global/safe-deployments/pull/228/commits/e50a93cdade1adf5e7dfa34e33dd164b25eefb48

package.json
mmv08: Why?

ademidun: I thought Richard wanted us to remove it.

> This should not be required as ts-node is a dev dependency

scripts/generateSummary.ts
```diff
@@ -0,0 +1,78 @@
...
+import { getSafeL2SingletonDeployment, getSafeSingletonDeployment } from '../src/safes';
+import { getProxyFactoryDeployment } from '../src/factories';
+import { getDefaultCallbackHandlerDeployment, getCompatibilityFallbackHandlerDeployment, getFallbackHandlerDeployment } from '../src/handler';
+import { getMultiSendDeployment, getMultiSendCallOnlyDeployment, getCreateCallDeployment, getSignMessageLibDeployment } from '../src/libs';
+import axios from 'axios';
+
+export async function generateSummary() {
+
+const response = await axios.get('https://chainid.network/chains.json');
+const ALL_CHAINS: any[] = response.data;
```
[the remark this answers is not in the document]
ademidun: I had `try-catch` before but removed it because if the request fails the entire script can't work. So there's really no `catch` behavior. 

try-catch could be useful if we could use a local`chains.json` file as a fallback in the `catch` portion.

@rmeissner would you be okay with that?

package.json
mmv08: I mentioned in the comment that it's not used anywhere, so it's an open question whether we want to keep it as a dependency. Since it is not related to the development of the package and is only used in the helper script to generate a table, I personally think it shouldn't be a dev dependency of the package.

PR #229, opened 2023-04-27
Add 1.4.0 deployments

mmv08:
This PR:
- Adds addresses of the deployed 1.4.0 contracts
- Adds `getSimulateTxAccessorDeployment` method. We had a JSON file with deployments, but there was no function to query it
- Updates tests to use mock data. Some tests were written with the assumption that the artifacts wont change


moisses89 (commented):
I couldn't find this file and address, should be this added on this PR?
`reusing "TokenCallbackHandler" at 0xeDCF620325E82e3B9836eaaeFdc4283E99Dd7562`

mmv08:
> I couldn't find this file and address, should be this added on this PR? `reusing "TokenCallbackHandler" at 0xeDCF620325E82e3B9836eaaeFdc4283E99Dd7562`

I don't know to be honest. It's a renamed `DefaultCallbackHandler`, which wasn't included in previous deployments either (besides 1.1.1). We don't use it in our production apps. We might want to add it for the 1.3.0 deployment too.

mmv08:
I've accidentally discovered that the tests fail. I will fix the tests and add a GitHub action to run tests on PR/push.

rmeissner (approved):

PR #237, opened 2023-05-22
🔧 Updated createCall deployments for 1.4.0

cairoeth:
Needed to deploy createCall to many chains with the address matching. Saw that @mmv08 deployed to some chains, but it was missing many of them. You can compare the deployments here: [0x9b35Af71d77eaf8d7e40252370304687390A1A52](https://blockscan.com/address/0x9b35Af71d77eaf8d7e40252370304687390A1A52) vs. [0xD8Ba6cF1756343D3171b25301bAA0719286f7155](https://blockscan.com/address/0xD8Ba6cF1756343D3171b25301bAA0719286f7155). It includes 30 chain deployments, both testnets, and mainnets. zkSync has different addresses than the rest because they use their own CREATE contract, but both mainnet and testnet match.

Happy to expand to more chains if needed.


mmv08:
Our contracts are deployed deterministically, and the expected address for the createcall contract is `0x9b35Af71d77eaf8d7e40252370304687390A1A52`. Expected addresses can be found in our [changelog](https://github.com/safe-global/safe-contracts/blob/e870f514ad34cd9654c72174d6d4a839e3c6639f/CHANGELOG.md)

PR #243, opened 2023-06-13
Add older Fantom and Fantom testnet

jenikd:
Contracts are deployed by Fantom Foundation and verified on Fantom mainnet and testnet. Can be checked with [ftmscan](https://ftmscan.com/).

Adding older V1.2.0 contracts to be able to convert them to V1.3.0 with Safe Wallet Web

Fantom mainnet chainID 250
Fantom testnet chainID 4002


mmv08:
> Adding older V1.2.0 contracts

Please move them to 1.2.0 folder, you added them to 1.1.1

HonzaDajc:
Hi @mmv08, I'v gnosis_safe.json to v1.2.0 folder. For rest contracts the ABI has not changed from v1.1.1 so that's why I put them in there. If I move all them to v1.2.0 folder, then I had to specify also rest of the networks there, right?

HonzaDajc:
Hi @mmv08, if I add files like proxy factory and other libs into 1.2.0 folder, then also other files like src/factories.ts and src/libs.ts have to be changed to be able to find 1.2.0 abi version. Do you want me to do this kind of change?

mmv08:
Why did you decide to use 1.2.0 for the deployment in the first place? Our services do not support it, as we skipped to 1.3.0 right away. Also the addresses seem to differ as the deterministic deployment wasn't used

HonzaDajc:
> Why did you decide to use 1.2.0 for the deployment in the first place? Our services do not support it, as we skipped to 1.3.0 right away. Also the addresses seem to differ as the deterministic deployment wasn't used

We deployed them in time, when 1.3.0 was still not released and also there was not deterministic deployment of the 1.2.0 in time of our deploy.
I know, that this state is not ideal and users had plenty time to upgrade their Safes to 1.3.0 via GUI.
From version 1.3.0 everything is OK and all contracts are deployed with deterministic addresses.

Uxio0:
> We deployed them in time, when 1.3.0 was still not released and also there was not deterministic deployment of the 1.2.0 in time of our deploy. I know, that this state is not ideal and users had plenty time to upgrade their Safes to 1.3.0 via GUI. From version 1.3.0 everything is OK and all contracts are deployed with deterministic addresses.

You can use those contracts, but we have a clear policy about using deterministic addresses, so if you want your network to be included in this repo you need to deploy the contracts in the right addresses

nlordell:
Closing as the PR has become stale.

PR #248, opened 2023-06-28
Feature: Switch v1.4.1 to released

mmv08:
This PR:
- Switches the released flag for 1.4.1 to true
- Adjusts tests that checked that the latest returned version was 1.3.0

### DIFF
--- a/src/__tests__/factories.test.ts
+++ b/src/__tests__/factories.test.ts
@@ -1,13 +1,13 @@
 import ProxyFactory130 from '../assets/v1.3.0/proxy_factory.json';
-import ProxyFactory111 from '../assets/v1.1.1/proxy_factory.json';
+import ProxyFactory141 from '../assets/v1.4.1/safe_proxy_factory.json';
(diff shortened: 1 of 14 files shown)


rmeissner (approved):
What impact does this have on the interface that use this (i.e. Safe Wallet)? Would they start creating Safes with the new version?

mmv08:
@rmeissner:
>We have a global constant (set to 1.3.0) that we filter by. However, [some places](https://github.com/safe-global/web-core/blob/dev/src/services/contracts/safeContracts.ts) just filter by network. We could ensure everywhere relevant uses the constant and/or the released flag though.

That's their latest update from 3 months ago. I will double check with the web team before releasing

liliya-soroka:
@mmv08 , did you check the impact with the mobile team too ?

katspaugh:
This is definitely a breaking change for us. Should have been a major version, not a patch version. I've fixed it in the linked PR for multisends. Need to check other contracts too.

PR #460, opened 2024-04-24
feat: bbchain deployments

kkgc00bouncebit:
Network Name: BounceBit Mainnet
RPC Url: https://fullnode-mainnet.bouncebitapi.com
Chain ID: 6001
Currency Symbol: BB
Block explorer URL: https://mainnet.bbscan.io/


src/assets/v1.4.1/compatibility_fallback_handler.json
```diff
@@ -41,7 +41,8 @@
     "444444": "0xfd0732Dc9E303f09fCEf3a7388Ad10A83459Ec99",
     "11155111": "0xfd0732Dc9E303f09fCEf3a7388Ad10A83459Ec99",
     "11155420": "0xfd0732Dc9E303f09fCEf3a7388Ad10A83459Ec99",
-    "666666666": "0xfd0732Dc9E303f09fCEf3a7388Ad10A83459Ec99"
+    "666666666": "0xfd0732Dc9E303f09fCEf3a7388Ad10A83459Ec99",
+    "6001": "0xbD2D2CCF8286895400aecFfb40547e4768997245"
```
mmv08, line 45: The address should match the other addresses in the file, and also could you please fix the sorting?

mmv08 (changes requested):

kkgc00bouncebit:
Because of our contract condition are unique and diff from others and we also cant't keep the same as the rest due to some reason. Let me know if the above works for u guys too.

src/assets/v1.4.1/compatibility_fallback_handler.json
kkgc00bouncebit: Because of our contract condition are unique and diff from others and we also cant't keep the same as the rest due to some reason. Let me know if the above works for u guys too.

mmv08: This is a collection of deterministic deployments and we cannot include contracts deployed non-deterministically here.

mmv08: @kkgc00bouncebit, may I close the issue, or will you fix it?

kkgc00bouncebit: @mmv08 Could you please deploy the safe-singleton-factory for me, thank you very mush. https://github.com/safe-global/safe-singleton-factory/issues/445

kkgc00bouncebit: @mmv08 Already fixed it, Please take a loot when you have some time.

mmv08 (approved):

PR #649, opened 2024-06-17
Github Review script

remedcu:
This PR adds a bash script to review new contract deployments quickly.

This is an attempt from Ecosystem ~friday~.

The bash script first checks the addresses to ensure they are the same as the `default addresses` in the JSON file. Finally checks the code hash in the new chain matches the intended code hash.

This requires a few things:
- Installation of `gh` (for fetching PR files) and `cast` (for RPC-related calls).
- RPC which can take at least 9 requests per minute.


bin/github-review.sh
```diff
@@ -0,0 +1,313 @@
...
+if [[ -n "$(git status --porcelain)" ]]; then
+    echo "ERROR: Dirty Git index, please commit all changes before continuing" 1>&2
+    exit 1
+fi
+if ! command -v gh &> /dev/null; then
+    echo "ERROR: Please install the 'gh' GitHub CLI" 1>&2
+    exit 1
+fi
+if ! command -v cast &> /dev/null; then
+    echo "ERROR: Please install the 'cast' Foundry (forge)" 1>&2
```
nlordell, line 33: nit: Grammar.

bin/github-review.sh
```diff
@@ -0,0 +1,313 @@
...
+    exit 1
+fi
+
+pr=0
+chainid=0
+rpc=0
+version=""
+# TODO: "1.0.0" "1.1.1" "1.2.0"
```
nlordell, line 41: These are no longer officially supported, so we shouldn't be adding deployments for older Safe versions. I.e. I don't think this is a "TODO".

bin/github-review.sh
```diff
@@ -0,0 +1,313 @@
...
+            exit 1
+        fi
+        chainid="$(cast chain-id --rpc-url $3)"
+        if [[ $chainid != $2 ]]; then
+            echo "ERROR: RPC $3 doesn't match chain ID $2" 1>&2
+            usage
+            exit 1
+        fi
+        rpc=$3
+        if [[ ! " ${availableVersion[@]} " =~ " ${4} " ]]; then
```
nlordell, line 63: This is a cleaver way for checking list inclusion! (For other less Bash-y ppl, `=~` is regex match). I wonder if `1.4_1` works (i.e. is `.` the regex "any character" or an actual '.')?

bin/github-review.sh
```diff
@@ -0,0 +1,313 @@
...
+
+# Getting default address based on the chain id
+
+# v1.3.0 & v1.4.1
+cfhCI=$(jq -r '.networkAddresses["'$chainid'"]' src/assets/v$version/compatibility_fallback_handler.json)
+ccCI=$(jq -r '.networkAddresses["'$chainid'"]' src/assets/v$version/create_call.json)
+mscoCI=$(jq -r '.networkAddresses["'$chainid'"]' src/assets/v$version/multi_send_call_only.json)
+msCI=$(jq -r '.networkAddresses["'$chainid'"]' src/assets/v$version/multi_send.json)
+smlCI=$(jq -r '.networkAddresses["'$chainid'"]' src/assets/v$version/sign_message_lib.json)
+staCI=$(jq -r '.networkAddresses["'$chainid'"]' src/assets/v$version/simulate_tx_accessor.json)
```
nlordell, line 117: I feel like this can be simplified to a loop over the JSON files, and have the equality check be done in `jq`.

bin/github-review.sh
```diff
@@ -0,0 +1,313 @@
...
+if [[ $version == "1.3.0" ]]; then
+    cfhCH="0x03e69f7ce809e81687c69b19a7d7cca45b6d551ffdec73d9bb87178476de1abf"
+    ccCH="0x8155d988823a4f6f1bcbc76a64af8e510c4ce68819290d43cf24956bd24dee82"
+    gsl2CH="0x21842597390c4c6e3c1239e434a682b054bd9548eee5e9b1d6a4482731023c0f"
+    gsCH="0xbba688fbdb21ad2bb58bc320638b43d94e7d100f6f3ebaab0a4e4de6304b1c2e"
+    mscoCH="0xa9865ac2d9c7a1591619b188c4d88167b50df6cc0c5327fcbd1c8c75f7c066ad"
+    msCH="0x0208282bd262360d0320862c5ac70f375f5ed3b9d89a83a615b4d398415bdc83"
+    pfCH="0x337d7f54be11b6ed55fef7b667ea5488db53db8320a05d1146aa4bd169a39a9b"
+    smlCH="0x3ac65dea3cc9dd0d7b7b800f834e3d73415b4e944bb94555c3e4a08fb137e918"
+    staCH="0xb3fb9763869f2c09a2ac5a425d2dd6060bf7ef46b3899049d71a711e71e00f04"
```
mmv08, line 211: Idea for the future: Ideally these would be included in the json file

mmv08 (approved):

bin/github-review.sh
mmv08: (i approve after addressing Nick's comments) good job!

nlordell: > Ideally these would be included in the json file

Agree - you could read them with `jq` from a JSON file:

```jsonc
{
  "v1.3.0": {
    "GnosisSafe": "0xbba688fbdb21ad2bb58bc320638b43d94e7d100f6f3ebaab0a4e4de6304b1c2e",
    "GnosisSafeL2": "0x21842597390c4c6e3c1239e434a682b054bd9548eee5e9b1d6a4482731023c0f",
    // ...
  },
  "v1.4.1": {
    "Safe": "0x1fe2df852ba3299d6534ef416eefa406e56ced995bca886ab7a553e6d0c5e1c4",
    "SafeL2": "0xb1f926978a0f44a2c0ec8fe822418ae969bd8c3f18d61e5103100339894f81ff",
    // ...
  }
}
```

Can be read with:

```bash
jq -r --arg v $version --arg c $contract '.[$v].[$c]' contract-code-hashes.json 
```

bin/github-review.sh
remedcu: Copilot: `" ${4} "`: This is the fourth argument passed to the script, surrounded by spaces and quotes to form a regular expression. The spaces ensure that the match must be a whole word. For example, if `${4}` is `2`, it will match `2` in the array, but not `20` or `12`.

bin/github-review.sh
remedcu: Will remove 👍🏾

bin/github-review.sh
remedcu: Updated 👍🏾

bin/github-review.sh
remedcu: Updated 👍🏾

bin/github-review.sh
```diff
@@ -0,0 +1,151 @@
...
+versionContracts=()
+if [[ $version == "1.3.0" ]]; then
+    versionContracts=("${oneThreeZeroContracts[@]}")
+elif [[ $version == "1.4.1" ]]; then
+    versionContracts=("${oneFourOneContracts[@]}")
+fi
+
+versionCodeHashes=()
+for contract in "${versionContracts[@]}"; do
+    versionCodeHashes+=($(jq -r '.["'$version'"].["'$contract'"]' src/assets/code-hashes.json))
```
nlordell, line 125: One disadvantage of doing it this way over using `--arg` like I suggested is that special characters might break the `jq` query this way.

bin/github-review.sh
```diff
@@ -0,0 +1,151 @@
...
+    exit 1
+fi
+
+pr=0
+chainid=0
+rpc=0
+version=""
+availableVersion=("1.3.0" "1.4.1")
+case $# in
+    4)
```
nlordell, line 43: nit: why not use something like:

```
if [[ "$#" -ne 4 ]]; then
    usage
    exit 1
fi
```

bin/github-review.sh
nlordell: TIL, from the Bash manual:

>  If any part of the pattern is quoted, the quoted portion is matched literally.  This means every character in the quoted portion matches itself, instead of having any special pattern matching meaning. If the pattern is stored in a shell variable, quoting the variable expansion forces the entire pattern to be matched literally.  Treat bracket expressions in regular expressions carefully, since normal quoting and pattern characters lose their meanings between brackets.

So, the fact that the expression is quotes makes it be matched literally and NOT like a regular expression (contrary to what I understood from Copilot). For example:

```
$ [[ "10" =~ 1. ]] && echo true || echo false
true
$ [[ "10" =~ "1." ]] && echo true || echo false
false
$ pattern='1.'
$ [[ "10" =~ $pattern ]] && echo true || echo false
true
$ [[ "10" =~ "$pattern" ]] && echo true || echo false
false
```

Bash is SO weird...

src/assets/code-hashes.json
```diff
@@ -0,0 +1,24 @@
+{
```
nlordell, line 1: nit: I would keep this in the `bin` assets. I don't know if we want to publish this with the package or not (my intuition says "no").

You can even embed it in the `bash` script:

```
JSON="$(cat <<JSON
{
  // ...
}
JSON
)"
```

bin/github-review.sh
```diff
@@ -0,0 +1,151 @@
...
+        exit 1
+        ;;
+esac
+
+echo "Verifying Deployment Asset"
+gh pr diff $pr --patch | git apply --include 'src/assets/**'
+
+# Getting default addresses
+oneThreeZeroFiles=("compatibility_fallback_handler" "create_call" "gnosis_safe_l2" "gnosis_safe" "multi_send_call_only" "multi_send" "proxy_factory" "sign_message_lib" "simulate_tx_accessor")
+oneFourOneFiles=("compatibility_fallback_handler" "create_call" "multi_send_call_only" "multi_send" "safe_l2" "safe_proxy_factory" "safe" "sign_message_lib" "simulate_tx_accessor")
```
nlordell, line 80: nit, I would just inline these... I don't think the named variable adds any value here.

bin/github-review.sh
```diff
@@ -0,0 +1,151 @@
...
+
+# Checking default addresses and the addresses created for that chain id
+for i in {0..8}; do
+    if [[ ${versionAddresses[$i]} != ${versionAddressesCI[$i]} ]]; then
+        echo "ERROR: "${versionFiles[$i]}" default address is not the same as the one created for the chain id" 1>&2
+        exit 1
+    fi
+done
+
+echo "Default addresses are correct"
```
nlordell, line 109:
```suggestion
echo "network addresses are correct"
```

bin/github-review.sh
```diff
@@ -0,0 +1,151 @@
...
+    versionCodeHashes+=($(jq -r '.["'$version'"].["'$contract'"]' src/assets/code-hashes.json))
+done
+
+# Now getting the codehash (keccak of bytecode) for the default addresses created for that chain id
+echo "Fetching codehashes from the chain"
+
+versionCodeHashesCI=()
+for i in {0..8}; do
+    versionCodeHashesCI+=($(cast keccak $(cast code ${versionAddressesCI[$i]} --rpc-url $rpc)))
+done
```
nlordell, line 134: If we want to not depend on assets, we can just fetch the mainnet code hash.

bin/github-review.sh
```diff
@@ -0,0 +1,151 @@
...
+
+if [[ $version == "1.3.0" ]]; then
+    versionFiles=("${oneThreeZeroFiles[@]}")
+elif [[ $version == "1.4.1" ]]; then
+    versionFiles=("${oneFourOneFiles[@]}")
+fi
+
+# Getting default addresses
+for file in "${versionFiles[@]}"; do
+    versionAddresses+=($(jq -r '.defaultAddress' "src/assets/v$version/$file.json"))
```
nlordell, line 92: One issue with this method is that it doesn't work for zkSync based networks or all deployments in v1.3.0 (there are two possible addresses there).

bin/github-review.sh
nlordell: We decided to add `codeHash` field to the existing deployments

src/assets/code-hashes.json
nlordell: We will NOT inline the JSON (leads to an unruly and large bash file, which isn't nice). Instead, we will add `codeHash` field to each deployment JSON.

bin/github-review.sh
remedcu: Done 👍🏾

bin/github-review.sh
remedcu: Updated 👍🏾

bin/github-review.sh
remedcu: Updated 👍🏾

nlordell (approved):

PR #654, opened 2024-06-19
Add Code Hashes To All Deployments

nlordell:
This PR adds code hashes to all deployment files. This is useful for a couple of reasons:

1. It makes it easier for external parties to verify the validity of the code at a specific address (even if it isn't at the expected `CREATE2` address)
2. It simplifies the `github-review.sh` script that @remedcu is working on in #649
3. Isn't "tech debt" like the `defaultAddress` field, as the code hash should be the same regardless of the deployment address (so it does not interfere with the plans to improve the deployment JSON format).

For posterity, the changes were generated with:

```fish
for f in src/assets/*/*.json
  set json "$(jq --arg h (jsonrpc $rpc eth_getCode (jq -r .defaultAddress $f) latest | hdwallet hex decode - | hdwallet hash data -) '{ defaultAddress, released, contractName, version, codeHash: $h, networkAddresses, abi }' $f)"
  echo "$json" > $f
end
```

Note that this makes a few formatting changes to some of the deployment files, but IMO this is a good thing (as we now have consistent formatting across all files :tada:).

### DIFF
--- a/src/__tests__/utilts.test.ts
+++ b/src/__tests__/utilts.test.ts
@@ -19,16 +19,18 @@ describe('utils.ts', () => {
         defaultAddress: '',
         version: '',
         abi: [],
-        networkAddresses: { "1": "0xbeef" },
+        networkAddresses: { "1": "0xbeefbeefbeefbeefbeefbeefbeefbeefbeefbeef" },
         contractName: '',
+        codeHash: '0x0000000000000000000000000000000000000000000000000000000000000000',
         released: false,
       };
       const testReleasedDeployment: SingletonDeployment = {
         defaultAddress: '',
         version: '',
         abi: [],
-        networkAddresses: { "1": "0xbeef" },
+        networkAddresses: { "1": "0xbeefbeefbeefbeefbeefbeefbeefbeefbeefbeef" },
         contractName: '',
+        codeHash: '0x0000000000000000000000000000000000000000000000000000000000000000',
         released: true, // Default filter value
       };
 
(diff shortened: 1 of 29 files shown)


remedcu (approved):

mmv08:
nit: can this be a separate json file?

reason: this package is used on the client size, and this (small but still) addition increases an already large filesize, and we didn't get codehash verification functionality from the clients so far

nlordell:
> Nit: can this be a separate JSON file?

Hmm, I see your point. reducing size of the JSON files can be nice for client side. Note that these files are extremely compressable:

```
$ git checkout main -- safe.json
$ cat safe.json | gzip -c | wc --bytes
1944
$ git checkout add-code-hashes -- safe.json
$ cat safe.json | gzip -c | wc --bytes
1999
```

So it adds ~50bytes of data to the file.

The downside is that we have to maintain multiple files per contract, which I'm not a fan of (you have two "databases" - i.e. JSON files - whose keys need to be in sync). From a "JSON file design standpoint", the deployment file feels like the right place to put it.

mmv08:
> > Nit: can this be a separate JSON file?
> 
> Hmm, I see your point. reducing size of the JSON files can be nice for client side. Note that these files are extremely compressable:
> 
> ```
> $ git checkout main -- safe.json
> $ cat safe.json | gzip -c | wc --bytes
> 1944
> $ git checkout add-code-hashes -- safe.json
> $ cat safe.json | gzip -c | wc --bytes
> 1999
> ```
> 
> So it adds ~50bytes of data to the file. As a non-web developer, I don't think this is a huge deal (but I can be wrong).
> 
> The downside is that we have to maintain multiple files per contract, which I'm not a fan of (you have two "databases" - i.e. JSON files - whose keys need to be in sync). From a "JSON file design standpoint", the deployment file feels like the right place to put it.

Ahh, 50 bytes is negligible then

akshay-ap (approved):

PR #665, opened 2024-06-26
Modernise NPM Project

nlordell:
This PR does a series of small project updates in order to mordernize it slighty:
1. It moves away from Yarn 1.x as it is deprecated
2. It uses the new ESLint configuration format ("flattened" configuration).
3. Update dependency versions
4. Lint the project (this fixed a bunch of formatting things).

I chose to use NPM instead of PNPM after dicussions with @mmv08 as it is the default package manager and we don't need any PNPM features.


.github/workflows/test.yml
```diff
@@ -9,9 +9,8 @@ jobs:
...
-      - uses: actions/checkout@v3
-      - name: Use Node.js 18
-        uses: actions/setup-node@v3
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
         with:
-          node-version: 18
-      - run: yarn && yarn test
+          node-version: 20
+      - run: npm ci && npm test
```
remedcu, line 16: Should we also add a step to check `npm run lint`? or maybe in a different workflow?

mmv08 (approved):

.github/workflows/test.yml
nlordell: Added

Issue #348, opened 2024-01-04
Improve the JSON file structure to support multiple addresses per network


mmv08:
The big downside of the current package's structure is that consumers have to pre-load the entire JSON file with all deployments to enable deployment lookup. While I don't think there's a workaround, the JSON structure could be optimized. For example, for 1.3.0 deployments, most deployments are just mapping a chain id to one of the known addresses. Instead, it could be optimized like this:
```
addresses:
  eip155: 0x1
  noneip155: 0x2
  zksync: 0x3
deployments:
  1: eip155
  2: noneip155
  3: zksync
  4: eip155
```

Additionally, we can move the ABI outside of the main JSON file, so consumers just wanting to use the addresses don't have to load it.

**EDIT 25.06.2024:**
After looking into optimizing the package size, I realised that the compression algorithm does a great job of compressing everything to a decent size (24kb). We could shave off a few kilobytes, but I think it doesn't have a significant impact.

So, I would suggest converting this issue into something the network and wallet teams need - supporting multiple singletons per network. I believe the proposed structure fits well still:

```
addresses:
  eip155: 0x1
  noneip155: 0x2
  zksync: 0x3
deployments:
  1: [eip155, noneip155]
  2: [noneip155, eip155]
  3: zksync
  4: eip155
```

So now the deployments will either be an array or a key from the addresses mapping. The array should be sorted by the date of deployment. The change can be done by adding a new function to avoid breaking changes.

mmv08:
> While I don't think there's a workaround, the JSON structure could be optimized.

A backend service / onchain lookup solution can also be explored

katspaugh:
This is a good idea. In the UI we check safe-deployments to determine supported Safes, and so does the protocol-kit. If each chain had several addresses we could be backward-compatible with both types of addresses for 1.3.0 most notably for chains like Optimism.

rmeissner:
Couldn't we also switch around that instead of having a file per contract with list of chain ids, we could have a json per chain id with an object of addresses

mmv08:
> Couldn't we also switch around that instead of having a file per contract with list of chain ids, we could have a json per chain id with an object of addresses

I'm curious how a bundler would help this. Theoretically if all of the deployments are stored as static assets on the client side that may help (so you could just lookup the file needed, e.g., `app.safe.global/static/deployments/chains/1.json`)

nlordell:
[...]

PR #668, opened 2024-06-27
Feature: Improve the JSON file structure to support multiple addresses per network

mmv08:
This PR implements https://github.com/safe-global/safe-deployments/issues/348. The following has been done:
1. The JSON assets have been updated to the new format to support multiple addresses. Now, the format is the following:
- **released**: `boolean`
  - Indicates if the contract is released.
- **contractName**: `string`
  - The name of the contract.
- **version**: `string`
  - The version of the contract.
- **codeHash**: `string`
  - The hash of the contract code.
- **addresses**: `object`
  - Contains the canonical addresses for the contract.
  - **type**: `address`
    - The canonical address of the contract.
- **networkAddresses**: `object`
  - Maps network IDs to the canonical addresses.
  - **ID**: `string` or `string[]`
- **abi**: `array`
  - The ABI (Application Binary Interface) of the contract, containing multiple function and event definitions.
2. The backward compatibility is implemented through JSON to JS object mapping through JS functions. As of now, the functions return the same structure as previous versions, except for the `addresses` object addition.
3. To ensure 2), I copied all the old JSON assets to the tests folder and compared them against the newly returned objects

**Important:**
This PR doesn't add any new functions to make use of the new JSON structure. I decided to bring this out for review and temperature check and if this. gets accepted, I'll implement the functions in a new PR


src/jsonMigrationScript130.ts
```diff
@@ -0,0 +1,138 @@
+import fs from 'fs';
```
mmv08, line 1: This is the migration script I wrote. I'll keep it here until the PR gets merged in case any new networks get added.

src/__tests__/assets/v1/v1.3.0/create_call.json
```diff
@@ -0,0 +1,323 @@
...
+  "contractName": "CreateCall",
+  "version": "1.3.0",
+  "codeHash": "0x8155d988823a4f6f1bcbc76a64af8e510c4ce68819290d43cf24956bd24dee82",
+  "addresses": {
+    "canonical": "0x7cbB62EaA69F79e6873cD1ecB2392971036cFAa4",
+    "eip155": "0xB19D6FFc2182150F8Eb585b79D4ABcd7C5640A9d",
+    "zksync": "0xcB8e5E438c5c2b45FbE17B02Ca9aF91509a8ad56"
+  },
+  "networkAddresses": {
+    "1": "0x7cbB62EaA69F79e6873cD1ecB2392971036cFAa4",
```
nlordell, line 13: Maybe a smaller subset makes more sense for tests (as discussed in stand-up).

nlordell:
Do we need the `released`? Isn't it `true` for all contracts? If so I would consider removing it.

Also, AFAIU, the 1.3.0 migration script found issues with certain networks. Can you document them in the PR as well?

src/jsonMigrationScript130.ts
nlordell: I was going to suggest to keep it in `scripts` directory so that the migration can be replayed in the future.

src/types.ts
```diff
@@ -1,12 +1,57 @@
...
+}
+
+interface AbiInput {
+  indexed?: boolean;
+  internalType: string;
+  name: string;
+  type: string;
+}
+
+interface AbiItem {
```
nlordell, line 25: I would be careful with changing the types. In particular it might break packages as it might not be usable with `ethers` for example (I don't know if it uses slightly stricter tagged union types that discriminate on the `type` field) and would leave the change out of the PR TBH.

src/types.ts
```diff
@@ -1,12 +1,57 @@
...
+
+export interface SingletonDeploymentJSON {
+  released: boolean;
+  contractName: string;
+  version: string;
+  codeHash: string;
+  networkAddresses: Record<string, string | string[]>;
+  addresses: Record<string, string>;
+  /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
+  abi: any[];
```
nlordell, line 15: Interesting that here we keep any but not in the other type.

nlordell (commented):
In general, LGTM, left some comments.

src/__tests__/assets/v1/v1.3.0/create_call.json
mmv08: Did we discuss that? 🙈 

Those files are 1:1 copies (with only an addition of the `addresses` mapping) of the pre-upgrade JSON assets. Using production assets gives me more confidence in the correctness of the package. I also think crafting the assets requires more effort.

src/types.ts
mmv08: I think you're right that this might be a breaking change. I reverted it for now, but I think we should include this in one of the future releases.

mmv08:
> Do we need the released? Isn't it true for all contracts? If so I would consider removing it.

Previously, we added new versions with the released flag set to false: https://github.com/safe-global/safe-deployments/pull/248. To be honest, I'm not 100% aware of the meaning behind the flag, but it seems handy to have an option of adding deployments of a new version before "releasing" it. No strong opinion.

mmv08:
> Also, AFAIU, the 1.3.0 migration script found issues with certain networks. Can you document them in the PR as well?

In such cases, I just referenced the old JSON file. That was suggested by @gjeanmart in [Slack](https://5afe.slack.com/archives/C04G9CPMR3J/p1719318348642669?thread_ts=1719313753.857399&cid=C04G9CPMR3J) and I agree with it:
> Finally, regarding deprecated testnets and dead chains, I would rather keep them because we never know, someone might want to revive them or could still be working on a fork internally.

src/jsonMigrationScript130.ts
mmv08: Why do we want to keep it? I don't think we need to use it after the PR merged.

src/__tests__/assets/v1/v1.3.0/create_call.json
mmv08: That shouldn't be the case, I fixed the tests to check that the old files are a subset of the new files.

Docs on the jest matcher I used: https://jestjs.io/docs/expect#tomatchobjectobject

nlordell (approved):

mmv08:
I actually tested the compression and the new format shaved 6KB off the package size

Issue #662, opened 2024-06-24
Add zkSync Support To Automated New Chain Checks


nlordell:
As we are automating new chain checks, we need additional zkSync support for the scripts.

remedcu:
Solved with #683

PR #683, opened 2024-07-04
Deployment Review Script

remedcu:
This PR improves the review cycle for the `safe-deployment` by checking the new deployments when a new PR is created for a particular chain.

This PR:
- instead of an `ISSUE_TEMPLATE`, introduced a PR Template which will ask for the required data from the PR creator, and help run the CI script.
- adds a CI review script, which succeeds if everything seems ok, else, it helps the PR creator to know what might be a problem with merging the PR in the CI logs.
- updates the `github-review` script, which checks for edge cases when a chain ID with the highest number is added.
- updates the test to accommodate the change in the JSON structure for the `codehash` (because `v.1.3.0` has multiple deployment types).
- updated the JSON structure to include address and code hash within a single object with the deployment type as the key.
- ZKSync support in CI mentioned in #662 (Ex: https://github.com/remedcu/safe-deployments/pull/8)

Remaining TODO's:
- There is a possibility that a user removes one of the deployment types (currently only applicable with `v1.3.0`) from the current list. This can't be detected with the current script.


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,17 @@
+## Add new chain
+
+Template to provide information about the new chain. Only add extra information in the bottom section. Ensure that the contracts are deployed on the chain, if not deploy with [safe-contracts](https://github.com/safe-global/safe-contracts). This entire paragraph can be deleted.
```
nlordell, line 3: Nit: 

```suggestion
> Template to provide information about the new chain. Only add extra information in the bottom section. Ensure that the contracts are deployed on the chain, if not deploy with [safe-contracts](https://github.com/safe-global/safe-contracts). This entire paragraph can be deleted.
```

makes it a bit more visually clear that this is not intended to be part of the PR template.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,17 @@
...
+- Chain_ID: 
+
+Provide RPC URL for the chain (should be able to query atleast 10 requests per minute for automatic PR check).
+- RPC_URL: 
+
+What version (1.3.0/1.4.1) of contracts are deployed? NOTE: One issue per version and one pull request per version. Do not add `v` infront of the version.
+- Contract_Version: 
+
+Relevant information:
+Add any other relevant information like blockexplorer, any annotation...
```
nlordell, line 17: I think templates have a "form" system - see https://github.com/safe-global/safe-singleton-factory/blob/main/.github/ISSUE_TEMPLATE/new_chain.yml

This _may_ apply to pull request templates, but I'm not 100% sure.

remedcu: I did try with form structure, but it only works for Issue, and didn't for PR Templates.

remedcu: Only `.md` and `.txt` are supported: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates#pull-request-templates

scripts/codehash.js
```diff
@@ -0,0 +1,6 @@
+const { exec } = require('child_process');
+
+const defaultAddress = process.env.DEFAULTADDRESS;
+const rpc = process.env.RPCURL;
+
+exec('echo $(cast keccak $(cast code ' + defaultAddress + ' --rpc-url ' + rpc + ') > codehash.txt)');
```
nlordell, line 6: I don't understand why this is an NPM script.

Also, if we use NPM scripts, lets use the `ethers` dependency that we already include instead of shelling out to an external tool.

remedcu: Mentioned in the PR Description:
> adds a JS script for checking the code hash of the contract as GitHub Actions have a limit on the length of the arguments.

src/__tests__/assets/v1/v1.0.0/gnosis_safe.json
```diff
@@ -3,7 +3,9 @@
   "released": true,
   "contractName": "GnosisSafe",
   "version": "1.0.0",
-  "codeHash": "0xe1f1593df76e69abc2d692792c80f329457551d5e83dde597546a1d58764da80",
+  "codeHash": {
+    "canonical": "0xe1f1593df76e69abc2d692792c80f329457551d5e83dde597546a1d58764da80"
+  },
   "addresses": {
     "canonical": "0xb6029EA3B2c51D09a50B53CA8012FeEB05bDa35A"
   },
```
nlordell, line 11: nit: I would change this. This is like representing an array of structs as arrays of fields. Something like:

```suggestion
  "deployments": {
    "canonical": {
      "address": "0xb6029EA3B2c51D09a50B53CA8012FeEB05bDa35A",
      "codeHash": "0xe1f1593df76e69abc2d692792c80f329457551d5e83dde597546a1d58764da80"
    }
  },
```

.github/workflows/review.yml
```diff
@@ -0,0 +1,45 @@
...
+
+  review-result:
+    runs-on: ubuntu-latest
+    needs: [review-check]
+    steps:
+    - uses: actions/checkout@v4
+    - name: Post Success Comment to the PR
+      if: needs.review-check.outputs.review-result == 'Network addresses & Code hashes are correct'
+      run: |
+        gh pr comment "$PR_NUMBER" --repo "$REPO" --body "✅ Success: Network addresses & Code hashes are correct"
```
nlordell, line 40: Nit: Don't post this in a PR comment - just make CI fail (it will show as a failed check).

src/__tests__/assets/v1/v1.0.0/gnosis_safe.json
remedcu: I'm ok with this change if @mmv08 doesn't have a problem 👍🏾

.github/workflows/review.yml
remedcu: My reasoning: if the dev is subscribed to this repo, he will get a message with this content, more specifically the "✅", easier to spot and faster to approve. Check success won't provide this way.

remedcu: My thought: if the dev is subscribed to this repo, he will get a message with this content, more specifically the "✅", easier to spot and faster to approve. Check success won't provide this way.

scripts/codehash.js
nlordell: > adds a JS script for checking the code hash of the contract as GitHub Actions have a limit on the length of the arguments.

This doesn't make sense to me :see_no_evil: - what limit were we hitting exactly?

My point still stands though - I wouldn't use cast and use Ethers.js directly here. Shelling is kind of unsafe especially since `rpc` is a parameter from an external untrusted source.

.github/workflows/review.yml
nlordell: But it will show as a red X here:

image

To me, this is super obvious.

scripts/codehash.js
nlordell: Oh, I see `cast keccak $(cast code ...)` is being replaced with `cast keccak 0x.....` for a very long `0x....`.

You can just solve this with piping though: `cast code ... | cast keccak` (at least, as long as `cast keccak` takes in hex from stdin).

nlordell: Looks like cast doesn't like the trailing newline in the code hex output. This can be fairly easily solved with `tr`:

```
cast code 0x41675C099F32341bf84BFc5382aF534df5C7461a --rpc-url https://eth.llamarpc.com | tr -d '\n' | cast keccak
```

.github/workflows/review.yml
remedcu: Updated 👍🏾

scripts/codehash.js
remedcu: Awesome! Updated!

.github/PULL_REQUEST_TEMPLATE.md
remedcu: Updated 👍🏾

.github/workflows/review.yml
```diff
@@ -27,19 +27,9 @@ jobs:
       id: review-check-result
       run: |
         script_result=$(bash bin/github-review.sh ${PR_NUMBER} | tail -1)
```
nlordell, line 29: This will fail for non-deployment related PRs (which happen) - we should make sure that this check is ONLY run on deployment related PRs.

bin/github-review.sh
```diff
@@ -155,9 +155,7 @@ for file in "${versionFiles[@]}"; do
     for deploymentType in "${deploymentTypes[@]}"; do
         defaultAddress=$(jq -r --arg t "$deploymentType" '.addresses[$t]' "$file")
         defaultCodeHash=$(jq -r --arg t "$deploymentType" '.codeHash[$t]' "$file")
-        DEFAULTADDRESS=$defaultAddress RPCURL=$rpc npm run codehash
-        networkCodeHash=$(cat codehash.txt)
-        rm codehash.txt
+        networkCodeHash=$(cast code $defaultAddress --rpc-url $rpc | tr -d '\n' | cast keccak)
```
nlordell, line 157:
```suggestion
        networkCodeHash=$(cast code "$defaultAddress" --rpc-url "$rpc" | tr -d '\n' | cast keccak)
```

.github/workflows/review.yml
remedcu: A check based on the deployment file path has been added. It will fail for this one as there are changes made in this.

bin/github-review.sh
remedcu: Updated 👍🏾

.github/workflows/review.yml
nlordell: Is there any advantage to doing it that way instead of using `path` for the triggers?

https://github.com/safe-global/safe-modules/blob/c403aff29bed6e3e326ab36c820094f2ada41c23/.github/workflows/ci_recovery.yml#L2-L5

Maybe not supported for `pull_request` triggers?

.github/workflows/review.yml
```diff
@@ -0,0 +1,42 @@
...
+      run: |
+        check_result=$(gh pr diff ${PR_NUMBER} --name-only | grep -E 'src/assets/.*/.*\.json' || exit 0)
+        if [ -n "$check_result" ]; then
+          echo "check_result=true" >> "$GITHUB_OUTPUT"
+        fi
+    - name: Check PR
+      if: steps.deployment-file-check.outputs.deployment-changed != 'true'
+      run: |
+        script_result=$(bash bin/github-review.sh ${PR_NUMBER} | tail -1)
+        if [ "$script_result" != "Network addresses & Code hashes are correct" ]; then
```
nlordell, line 37: Currently, it hides the script stdout - why not just `exit 1` in the script on error? Feels like the more "natural" way to communicate failure in Bash.

bin/github-review.sh
```diff
@@ -45,53 +41,130 @@ if ! [[ $1 =~ ^[0-9]+$ ]]; then
...
-    echo "ERROR: $2 is not a valid Chain ID number" 1>&2
+prChainID="$(gh pr view $pr | sed -nE 's/.*Chain_ID: ([0-9]+).*/\1/p')"
+if ! [[ $prChainID =~ ^[0-9]+$ ]]; then
+    echo "ERROR: $prChainID is not a valid Chain ID number" 1>&2
     usage
     exit 1
 fi
-rpc=$3
+chainlistURL="https://chainlist.org/chain/$prChainID"
+if [[ "$(curl -s "$chainlistURL")" == 'nope' ]]; then
```
nlordell, line 51: This kind of check is very brittle - if Chainlist changes how they format errors, it will break.

I recommend:

```suggestion
chainlistURL="https://chainlist.org/chain/$prChainID"
if ! curl --fail -s -o /dev/null "$chainlistURL"; then
```

This uses `--fail` so `curl` exist with a non-0 error code if there is an HTTP error and doesn't rely on how Chainlist present HTTP errors (which is non-standard). We also send the output to `/dev/null` because we don't care about it.

Note that, depending on the chainlist page, it could also run into the "too long" errors you had with the cast command.

bin/github-review.sh
```diff
@@ -45,53 +41,130 @@ if ! [[ $1 =~ ^[0-9]+$ ]]; then
...
+    echo "ERROR: Chainlist URL $chainlistURL doesn't exist" 1>&2
+    usage
+    exit 1
+fi
+rpc="$(gh pr view $pr | sed -nE 's/.*RPC_URL: (https?:\/\/[^ ]+).*/\1/p')"
 chainid="$(cast chain-id --rpc-url $rpc)"
```
nlordell, line 57:
```suggestion
chainid="$(cast chain-id --rpc-url "$rpc")"
```

bin/github-review.sh
```diff
@@ -45,53 +41,130 @@ if ! [[ $1 =~ ^[0-9]+$ ]]; then
...
     exit 1
 fi
-rpc=$3
+chainlistURL="https://chainlist.org/chain/$prChainID"
+if [[ "$(curl -s "$chainlistURL")" == 'nope' ]]; then
+    echo "ERROR: Chainlist URL $chainlistURL doesn't exist" 1>&2
+    usage
+    exit 1
+fi
+rpc="$(gh pr view $pr | sed -nE 's/.*RPC_URL: (https?:\/\/[^ ]+).*/\1/p')"
```
nlordell, line 56: nit: stricter parsing here:
```suggestion
rpc="$(gh pr view $pr | sed -nE 's/^- RPC_URL: (https?:\/\/[^ ]+).*/\1/p')$"
```

bin/github-review.sh
```diff
@@ -45,53 +41,130 @@ if ! [[ $1 =~ ^[0-9]+$ ]]; then
...
 chainid="$(cast chain-id --rpc-url $rpc)"
-if [[ $chainid != $2 ]]; then
-    echo "ERROR: RPC $rpc doesn't match chain ID $2" 1>&2
+if [[ $chainid != $prChainID ]]; then
+    echo "ERROR: RPC $rpc doesn't match chain ID $prChainID" 1>&2
     usage
     exit 1
 fi
-version=$4
+version="$(gh pr view $pr | sed -nE 's/.*Contract_Version: (1\.[3-4]\.[0-1]).*/\1/p')"
```
nlordell, line 63: Idem on stricter parsing here.

bin/github-review.sh
```diff
@@ -45,53 +41,130 @@ if ! [[ $1 =~ ^[0-9]+$ ]]; then
...
+    echo "ERROR: PR contains changes in files other than src/assets/v$version/*.json" 1>&2
+    echo "Changed files:"
+    echo "$(gh pr diff $pr --name-only | grep -v -e 'src/assets/v'$version'/.*\.json')" 1>&2
+    exit 1
+fi
+
+echo "Checking changes to assets files"
+fileLineChangeJSON=$(gh pr view $pr --json files) # This line fetches the JSON output of the files changed in the PR
+edgeCase=0
+echo "$fileLineChangeJSON" | jq -r '.files[] | "\(.path) \(.additions) \(.deletions)"' | while read -r line; do
```
nlordell, line 83: This is very cool.

bin/github-review.sh
```diff
@@ -45,53 +41,130 @@ if ! [[ $1 =~ ^[0-9]+$ ]]; then
...
+    path=$(echo $line | cut -d ' ' -f1)
```
nlordell, line 84: uber-nano-nit: If `path` has a space, this will break. I suggest using a non-standard character (`\0` is commonly used).

Not a huge deal for our case, since we know the file names - but just wanted to point it out in case it is easy to fix (if it isn't - leave it as is).

bin/github-review.sh
```diff
@@ -45,53 +41,130 @@ if ! [[ $1 =~ ^[0-9]+$ ]]; then
...
+    additions=$(echo $line | cut -d ' ' -f2)
+    deletions=$(echo $line | cut -d ' ' -f3)
+
+    # Now you can perform checks on $additions and $deletions as per your requirements
+    if [[ ($additions == 2 && $deletions == 1) ]]; then
```
nlordell, line 89: Always `"` in if statements - there are some real rough edges when `$additions` is empty for example.

```suggestion
    if [[ ("$additions" == 2 && "$deletions" == 1) ]]; then
```

bin/github-review.sh
```diff
@@ -45,53 +41,130 @@ if ! [[ $1 =~ ^[0-9]+$ ]]; then
...
+if [[ -n "$(gh pr diff $pr --name-only | grep -v -e 'src/assets/v'$version'/.*\.json')" ]]; then
+    echo "ERROR: PR contains changes in files other than src/assets/v$version/*.json" 1>&2
+    echo "Changed files:"
+    echo "$(gh pr diff $pr --name-only | grep -v -e 'src/assets/v'$version'/.*\.json')" 1>&2
+    exit 1
+fi
+
+echo "Checking changes to assets files"
+fileLineChangeJSON=$(gh pr view $pr --json files) # This line fetches the JSON output of the files changed in the PR
+edgeCase=0
```
nlordell, line 82: nit: this is a very ambiguous variable name IMO. Why not:

```suggestion
MaybeHighestChainId=0
```

bin/github-review.sh
```diff
@@ -45,53 +41,130 @@ if ! [[ $1 =~ ^[0-9]+$ ]]; then
...
+    fi
+done
+
+if [[ $edgeCase == 1 ]]; then
+    echo "Edge case when adding chain with the highest chain id number"
+    diffPatchSeparated=($(gh pr diff $pr | grep -E '^[+-] '))
+    # Adding three elements at a time together to compare from the output array
+    diffPatch=()
+    for ((i=0; i<${#diffPatchSeparated[@]}; i+=3)); do
+        diffPatch+=("${diffPatchSeparated[i]} ${diffPatchSeparated[i+1]} ${diffPatchSeparated[i+2]}")
```
nlordell, line 103: Doesn't this depend on how many before lines and after lines are included in the diff?

bin/github-review.sh
```diff
@@ -45,53 +41,130 @@ if ! [[ $1 =~ ^[0-9]+$ ]]; then
...
+    if [[ ($additions == 2 && $deletions == 1) ]]; then
+        edgeCase=1
+    elif [[ ($additions != 1 || $deletions != 0) ]]; then
+        echo "ERROR: $path has invalid changes" 1>&2
+        exit 1
+    fi
+done
+
+if [[ $edgeCase == 1 ]]; then
+    echo "Edge case when adding chain with the highest chain id number"
```
nlordell, line 98: This loop is quite complicated to me, and I'm not sure I understand it. I will try with fresh eyes on Monday.

bin/github-review.sh
```diff
@@ -45,53 +41,130 @@ if ! [[ $1 =~ ^[0-9]+$ ]]; then
...
+fi
+
 echo "Verifying Deployment Asset"
-gh pr diff $pr --patch | git apply --include 'src/assets/**'
+gh pr diff $pr --patch | git apply --include 'src/assets/v'$version'/**'
 
 # Getting default addresses, address on the chain and checking code hash.
+deploymentTypes=($(jq -r --arg c "$chainid" '.networkAddresses[$c]' "$versionFiles"))
+if [[ $deploymentTypes == "[" ]]; then
+    deploymentTypes=($(jq -r --arg c "$chainid" '.networkAddresses[$c][]' "$versionFiles"))
```
nlordell, line 151: These should not be arrays.

nlordell: Yes it should.

bin/github-review.sh
```diff
@@ -45,53 +41,130 @@ if ! [[ $1 =~ ^[0-9]+$ ]]; then
...
+        ((counter++))
+    done
+fi
+
 echo "Verifying Deployment Asset"
-gh pr diff $pr --patch | git apply --include 'src/assets/**'
+gh pr diff $pr --patch | git apply --include 'src/assets/v'$version'/**'
 
 # Getting default addresses, address on the chain and checking code hash.
+deploymentTypes=($(jq -r --arg c "$chainid" '.networkAddresses[$c]' "$versionFiles"))
```
nlordell, line 149: IMO this is easier to express in `jq`:

deploymentTypes=($(jq -r --arg c "$chainid" '[.networkAddresses[$c]] | flatten | .[]' "$versionFiles"))

bin/github-review.sh
```diff
@@ -45,53 +41,130 @@ if ! [[ $1 =~ ^[0-9]+$ ]]; then
...
+if [[ $deploymentTypes == "[" ]]; then
+    deploymentTypes=($(jq -r --arg c "$chainid" '.networkAddresses[$c][]' "$versionFiles"))
+fi
```
nlordell, line 152: IMO this is easier to express in `jq`:

```suggestion
deploymentTypes=($(jq -r --arg c "$chainid" '[.networkAddresses[$c]] | flatten | .[]' "$versionFiles"))
```

bin/github-review.sh
```diff
@@ -45,53 +41,130 @@ if ! [[ $1 =~ ^[0-9]+$ ]]; then
...
-        echo "ERROR: "$file" default address is not the same as the one created for the chain id" 1>&2
-        exit 1
-    fi
-    defaultCodeHash=$(jq -r '.codeHash' "$file")
-    networkCodeHash=$(cast keccak $(cast code $networkAddress --rpc-url $rpc))
-    if [[ $defaultCodeHash != $networkCodeHash ]]; then
-        echo "ERROR: "$file" code hash is not the same as the one created for the chain id" 1>&2
-        exit 1
-    fi
+    for deploymentType in "${deploymentTypes[@]}"; do
```
nlordell, line 154: Doesn't this run for each version type for each file for each file? My understanding is that the `deploymentTypes` is built for **all** files (so it will have a duplicate for each file).

bin/github-review.sh
```diff
@@ -45,53 +41,130 @@ if ! [[ $1 =~ ^[0-9]+$ ]]; then
...
 git restore --ignore-unmerged -- src/assets
 
 # NOTE/TODO
-# - We should still manually verify there is no extra code in the PR.
-# - We could fetch the version, chain id and rpc from the PR (needs a standard format and possibly a tag in PR).
-# - We can approve PR using Github CLI. Should only be added after all manual tasks can be automated.
-# - Supporting zkSync and alternative deployment addresses for 1.3.0 contracts.
-# - If there are changes in any other path other than src/assets, the script should show an error.
-# - We should ensure there are not more than a single chain ID being added in a PR.
+# - We should still manually verify there is no removal of deployment types for a single chain.
```
nlordell, line 170: I'm not sure I understand (it late on a Friday).

.github/workflows/review.yml
remedcu: I just checked, it does support `pull_request` triggers. Updated 👍🏾

.github/workflows/review.yml
remedcu: Updated 👍🏾

bin/github-review.sh
remedcu: Nice suggestion, updated. TIL about `--fail`.

bin/github-review.sh
remedcu: Done 👍🏾

bin/github-review.sh
remedcu: Done 👍🏾

bin/github-review.sh
remedcu: Done 👍🏾

bin/github-review.sh
remedcu: In this case, `additions` won't be empty. Either it will be a zero or non-zero value based on the changes for that particular file. For consistency, made the change 👍🏾

bin/github-review.sh
remedcu: Yes, but we already know that if it enters this `if` branch, there are only 2 additions and 1 subtraction. With `grep` we only take the actual lines added and deleted in the diff patch. And if it doesn't follow the intended pattern (based on the check we have written), then it errors out, as it should be.

bin/github-review.sh
remedcu: Yes, it will run for each version if there is more than one version deployed for that particular chain (only limited to `v1.3.1`). Which I think should not be a problem, other than if the RPC have some limit of less than 18 requests per minute or so.

bin/github-review.sh
remedcu: Currently, if there are two versions mentioned for a chain, and a malicious person deletes one deployment type from all files. It still will pass the current review script. Thus, as a next step of improvement, we should be able to detect if a chain is removed or not from the existing list (only applicable to `v1.3.1` for now).

bin/github-review.sh
remedcu: Updated, lot of new things to learn with this bash script ❤️

remedcu:
Test results:
- Unrelated changes: https://github.com/remedcu/safe-deployments/pull/7
- ZKSync: https://github.com/remedcu/safe-deployments/pull/8 (Seems the problem was something based on the change we made, it now works)
- Canonical & EIP-155: https://github.com/remedcu/safe-deployments/pull/9
- EIP-155: https://github.com/remedcu/safe-deployments/pull/10
- Canonical: https://github.com/remedcu/safe-deployments/pull/11
- Highest Chain Addition: https://github.com/remedcu/safe-deployments/pull/12

In all the PR's the `test` CI fail, because to check things, I deleted some Chains. Before making this PR "Ready for review", will restore it.

bin/github-review.sh
remedcu: Did you mean something like:
```
    path=$(echo $line | cut -d '\0' -f1)
```

In this case, as you mentioned, as file names are known, this might not be a scenario for us.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,17 @@
...
+> Template to provide information about the new chain. Only add extra information in the bottom section. Ensure that the contracts are deployed on the chain, if not deploy with [safe-contracts](https://github.com/safe-global/safe-contracts). This entire paragraph can be deleted.
+
+Please fill the following form:
+
+Provide the Chain ID (Only 1 chain id per PR).
+- Chain_ID: 
+
+Provide RPC URL for the chain (should be able to query atleast 10 requests per minute for automatic PR check).
+- RPC_URL: 
+
```
nlordell, line 12: I wish there was a way to get this programmatically from Chainlist... My concern is that someone can put in an RPC for an "unofficial" node that lies and gives false sense of security with the :heavy_check_mark: in CI.

Is there a way to programatically get this somehow? I tried snooping on Chainlist but didn't find anything.

I'm OK with it as-is for now, but something to potentially improve in the future.

src/types.ts
```diff
@@ -74,11 +72,13 @@ export interface SingletonDeploymentV2 {
   released: boolean;
   contractName: string;
   version: string;
-  codeHash: string;
+  // The address & hash of the contract code, where the key is the deployment type.
+  // There could be multiple deployment types: canonical, eip155, zksync
+  // Ex: deployments: { "canonical": { "codeHash": "0x1234", "address": "0x5678"}}
```
nlordell, line 77: nit: I wouldn't duplicate the comment here.

.github/PULL_REQUEST_TEMPLATE.md
remedcu: There could be something we can write in bash to extract the RPC from the list on the Chainlist website, but it will be highly dependent on their website structure. We could have a list of things which we can improve on the script as a future iteration, and this could be one of them.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,17 @@
...
+What version (1.3.0/1.4.1) of contracts are deployed? NOTE: One issue per version and one pull request per version. Do not add `v` infront of the version.
+- Contract_Version:
```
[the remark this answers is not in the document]
remedcu: Nice idea indeed!

bin/github-review.sh
nlordell: This can be fixed by deduping `deploymentTypes`. My point is that, AFAIU from how JQ works with multiple files specified, for 1.3.1 it will be something like:

```
deploymentTypes=(canonical eip155 canonical eip155 canonical eip155 canonical eip155 canonical eip155 canonical eip155 canonical eip155 canonical eip155 canonical eip155)
```

bin/github-review.sh
```diff
@@ -45,53 +41,127 @@ if ! [[ $1 =~ ^[0-9]+$ ]]; then
...
+        ((counter++))
+    done
+fi
+
 echo "Verifying Deployment Asset"
-gh pr diff $pr --patch | git apply --include 'src/assets/**'
+gh pr diff $pr --patch | git apply --include 'src/assets/v'$version'/**'
 
 # Getting default addresses, address on the chain and checking code hash.
+deploymentTypes=($(jq -r --arg c "$chainid" '[.networkAddresses[$c]] | flatten | .[]' "$versionFiles"))
```
nlordell, line 149: I think this is dangerous for arrays - you should use:
```suggestion
deploymentTypes=($(jq -r --arg c "$chainid" '[.networkAddresses[$c]] | flatten | .[]' "${versionFiles[@]}"))
```

remedcu: Now [this comment](https://github.com/safe-global/safe-deployments/pull/683#discussion_r1668565166) make more sense. So, instead of all the version files, I am intentionally only selecting a single file (defaults to the first value in the array), as the deployment types should be the same for all the files.

So, if I make it to parse the values from all files (based on the above suggestion), then it will return all the deployment types of all the files, which is not the intended output.

.github/PULL_REQUEST_TEMPLATE.md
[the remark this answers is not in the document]
remedcu: Added to TODO 👍🏾

.github/PULL_REQUEST_TEMPLATE.md
remedcu: Added to TODO 👍🏾

bin/github-review.sh
nlordell: Oh... I guess I don't know bash arrays very well. Is `"$versionFiles"` implicitly doing `"${versionFiles[0]}"`?

remedcu: I believe yes: https://stackoverflow.com/a/47368011/7520013

.github/PULL_REQUEST_TEMPLATE.md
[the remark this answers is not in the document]
remedcu: Scratched the TODO, completed in the current one. Now just the Chain ID and RPC is required 👍🏾

src/types.ts
```diff
@@ -74,11 +72,10 @@ export interface SingletonDeploymentV2 {
   released: boolean;
   contractName: string;
   version: string;
-  codeHash: string;
+  deployments: Record<string, Record<string, string>>;
```
mmv08, line 75: This type can be more specific
```suggestion
  deployments: Record<AddressType, Record<'address' | 'codeHash', string>>;
```

remedcu: Updated 👍🏾

src/__tests__/assets/v1/v1.3.0/multi_send.json
```diff
@@ -3,11 +3,19 @@
...
-    "eip155": "0x998739BFdAAdde7C933B942a68053933098f9EDa",
-    "zksync": "0x0dFcccB95225ffB03c6FBB2559B530C2B7C8A912"
+  "deployments": {
+    "canonical": {
+      "address": "0xA238CBeb142c10Ef7Ad8442C6D1f9E89e07e7761",
+      "codeHash": "0x0208282bd262360d0320862c5ac70f375f5ed3b9d89a83a615b4d398415bdc83"
+    },
+    "eip155": {
+      "address": "0x998739BFdAAdde7C933B942a68053933098f9EDa",
+      "codeHash": "0x81db0e4afdf5178583537b58c5ad403bd47a4ac7f9bde2442ef3e341d433126a"
```
mmv08, line 13: Interesting that the codehash is different here 🤔

mmv08 (approved):
gj 🔥

src/__tests__/assets/v1/v1.3.0/multi_send.json
remedcu: The `codeHash` for `multiSend` and `SimulateTxAccessor` is different as it uses the `address(this)` within the constructor, thus making a different bytecode based on the address.

witaliy12 (approved):

PR #684, opened 2024-07-04
Feature: Multiple deployments functions

### LINKED ISSUE #685: Implement New Accessor Functions for Deploments with Multiple Addresses
Now that the JSON has been updated to allow for multiple addresses per network, new accessor functions for fetching deployments including multiple addresses need to be added to the repository.

Work has already begun in #684

mmv08:
This PR is a new API proposal for the functions that support the updated JSON format, which supports multiple deployments per network. Implements https://github.com/safe-global/safe-deployments/issues/685

The complexity of supporting multiple formats is handled by the utility function `findDeployments`, which contains a function overload with the deployment format parameter.

The new functions are added alongside the [existing API](https://github.com/safe-global/safe-deployments?tab=readme-ov-file#usage) functions but with plural `deployments` at the end of the function name.

**Example function:**
```
export const getProxyFactoryDeployments = (filter?: DeploymentFilter): SingletonDeploymentV2 | undefined => {
  return findDeployment(filter, factoryDeployments, DeploymentFormats.MULTIPLE);
};
```

**My reason for going with this approach**:
All the complexity is kept within the internal `findDeployment` function. Developers do not have to import and understand which format means what and don't have to import the additional `DeploymentFilter` enum. Type shenanigans are kept to a minimum. Also, the alternative approach below would require refactoring package functions to be defined via the `function` keyword because arrow functions do not support function overloads properly: https://www.perplexity.ai/search/typescript-arrow-function-over-tit1.0R8Rracx_PAUr1nVw#0

**Alternative approach:**
Add function overloads to the existing functions like this:
```
export const getSafeSingletonDeployment = (filter?: DeploymentFilter, format = DeploymentFormats.SINGLETON) => {
  return findDeployment(filter, _safeDeployments, format);
};
```

### DIFF
--- a/src/deployments.ts
+++ b/src/deployments.ts
@@ -1,0 +1,90 @@
+import { SingletonDeploymentJSON } from './types';
+
+// This is a file where all the deployments are consolidated
+// We do it in a separate file so we don't have to repeat comments about the array order and the type casting.
+// We use some specific types (like `AddressType`) in the `SingletonDeploymentJSON` type, but the TypeScript cannot infer that from the JSON files.
+// So we need to cast them to `SingletonDeploymentJSON` manually. The casting is valid because we have a test in `__tests__/assets.test.ts` that checks that the JSON files are valid.
+// The arrays are sorted by preference, which at the moment means from the most recent version to the oldest.
+// The arrays are prefixed with an underscore because they are not meant to be imported directly.
+
+import SimulateTxAccessor130 from './assets/v1.3.0/simulate_tx_accessor.json';
+import SimulateTxAccessor141 from './assets/v1.4.1/simulate_tx_accessor.json';
+
+const _ACCESSOR_DEPLOYMENTS = [SimulateTxAccessor141, SimulateTxAccessor130] as SingletonDeploymentJSON[];
+
+import ProxyFactory100 from './assets/v1.0.0/proxy_factory.json';
+import ProxyFactory111 from './assets/v1.1.1/proxy_factory.json';
+import ProxyFactory130 from './assets/v1.3.0/proxy_factory.json';
+import SafeProxyFactory141 from './assets/v1.4.1/safe_proxy_factory.json';
+
+const _FACTORY_DEPLOYMENTS = [
+  SafeProxyFactory141,
+  ProxyFactory130,
+  ProxyFactory111,
+  ProxyFactory100,
+] as SingletonDeploymentJSON[];
+
+import DefaultCallbackHandler130 from './assets/v1.1.1/default_callback_handler.json';
+
+const _DEFAULT_CALLBACK_HANDLER_DEPLOYMENTS = [DefaultCallbackHandler130] as SingletonDeploymentJSON[];
+
+import CompatibilityFallbackHandler130 from './assets/v1.3.0/compatibility_fallback_handler.json';
+import CompatibilityFallbackHandler141 from './assets/v1.4.1/compatibility_fallback_handler.json';
+
+const _COMPAT_FALLBACK_HANDLER_DEPLOYMENTS = [
+  CompatibilityFallbackHandler141,
+  CompatibilityFallbackHandler130,
+] as SingletonDeploymentJSON[];
+
+import Safe141 from './assets/v1.4.1/safe.json';
+import GnosisSafe130 from './assets/v1.3.0/gnosis_safe.json';
+import GnosisSafe120 from './assets/v1.2.0/gnosis_safe.json';
+import GnosisSafe111 from './assets/v1.1.1/gnosis_safe.json';
+import GnosisSafe100 from './assets/v1.0.0/gnosis_safe.json';
+
+const _SAFE_DEPLOYMENTS = [
+  Safe141,
+  GnosisSafe130,
+  GnosisSafe120,
+  GnosisSafe111,
+  GnosisSafe100,
+] as SingletonDeploymentJSON[];
+
+import SafeL2141 from './assets/v1.4.1/safe_l2.json';
+import GnosisSafeL2130 from './assets/v1.3.0/gnosis_safe_l2.json';
+
+const _SAFE_L2_DEPLOYMENTS = [SafeL2141, GnosisSafeL2130] as SingletonDeploymentJSON[];
+
+import MultiSend111 from './assets/v1.1.1/multi_send.json';
+import MultiSend130 from './assets/v1.3.0/multi_send.json';
+import MultiSend141 from './assets/v1.4.1/multi_send.json';
+
+const _MULTI_SEND_DEPLOYMENTS = [MultiSend141, MultiSend130, MultiSend111] as SingletonDeploymentJSON[];
+
+import MultiSendCallOnly130 from './assets/v1.3.0/multi_send_call_only.json';
+import MultiSendCallOnly141 from './assets/v1.4.1/multi_send_call_only.json';
+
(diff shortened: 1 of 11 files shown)


nlordell (approved):
LGTM.

mmv08:
I put this on pause because of a conflict with https://github.com/safe-global/safe-deployments/pull/683. I'll let @remedcu make the final changes to the JSON file structure. Also, in the meantime, we can wait for feedback from @gjeanmart and @katspaugh

src/__tests__/libs.test.ts
```diff
@@ -1,39 +1,78 @@
...
 } from '../libs';
+// @ts-expect-error The function is conditionally exported only if the process.env.NODE_ENV is 'test'. TypeScript will not be able to figure it out.
+import { mapJsonToDeploymentsFormatV2 } from '../utils';
 
 describe('libs.ts', () => {
   describe('getMultiSendDeployment', () => {
     it('should find the preferred deployment first', () => {
       const result = getMultiSendDeployment();
       expect(result).toMatchObject(MultiSend141);
-      expect(result).not.toBe(MultiSend111);
```
mmv08, line 18: I removed this because it seemed redundant. Same reasoning for other similar removes

src/__tests__/safes.test.ts
```diff
@@ -1,10 +1,19 @@
...
 import GnosisSafe100 from './assets/v1/v1.0.0/gnosis_safe.json';
-import { getSafeSingletonDeployment, getSafeL2SingletonDeployment } from '../safes';
+import {
+  getSafeSingletonDeployment,
+  getSafeL2SingletonDeployment,
+  getSafeSingletonDeployments,
+  getSafeL2SingletonDeployments,
+} from '../safes';
+// @ts-expect-error The function is conditionally exported only if the process.env.NODE_ENV is 'test'. TypeScript will not be able to figure it out.
+import { mapJsonToDeploymentsFormatV2 } from '../utils';
```
nlordell, line 16: uber-nit: To abstract this away, you could have a `src/__tests__/utils.ts` file that re-exports this symbol, so you only have to work around and document this TS issue in a single plance, and all tests import from:

```suggestion
import { mapJsonToDeploymentsFormatV2 } from './utils';
```

src/handler.ts
```diff
@@ -1,35 +1,81 @@
...
+/**
+ * Get the default callback handler deployment based on the provided filter.
+ * @param {DeploymentFilter} [filter] - Optional filter to apply to the deployment search.
+ * @returns {SingletonDeployment | undefined} - The found deployment or undefined if not found.
+ */
 export const getDefaultCallbackHandlerDeployment = (filter?: DeploymentFilter): SingletonDeployment | undefined => {
   return findDeployment(filter, defaultCallbackHandlerDeployments);
 };
 
+export const getDefaultCallbackHandlerDeployments = (filter?: DeploymentFilter): SingletonDeploymentV2 | undefined => {
```
nlordell, line 25: Should this also have a NatSpec function (saw you adding them in other places)?

nlordell (approved):
LGTM. 2 small things, but approving to reduce back-and-forth:

- One suggestion to avoid the long `@ts-expect-error` explanation in multiple places (but feel free to merge without this suggestion if you don't agree)
- One missing function documentation

mmv08:
@nlordell @remedcu I pushed some changes after the updated script from Shebin, namely:
- exported all the deployments to the `deployments.ts` file (reasons are outlined in the comments in the file itself)
- improved types

src/deployments.ts
```diff
@@ -0,0 +1,90 @@
+import { SingletonDeploymentJSON } from './types';
+
+// This is a file where all the deployments are consolidated
+// We do it in a separate file so we don't have to repeat comments about the array order and the type casting.
+// We use some specific types (like `AddressType`) in the `SingletonDeploymentJSON` type, but the TypeScript cannot infer that from the JSON files.
+// So we need to cast them to `SingletonDeploymentJSON` manually. The casting is valid because we have a test in `__tests__/assets.test.ts` that checks that the JSON files are valid.
+// The arrays are sorted by preference, which at the moment means from the most recent version to the oldest.
+// The arrays are prefixed with an underscore because they are not meant to be imported directly.
```
nlordell, line 8: This shouldn't affect a bundler's ability to tree-shake right?

src/deployments.ts
```diff
@@ -0,0 +1,90 @@
...
+
+const _CREATE_CALL_DEPLOYMENTS = [CreateCall141, CreateCall130] as SingletonDeploymentJSON[];
+
+import SignMessageLib130 from './assets/v1.3.0/sign_message_lib.json';
+import SignMessageLib141 from './assets/v1.4.1/sign_message_lib.json';
+
+const _SIGN_MESSAGE_LIB_DEPLOYMENTS = [SignMessageLib141, SignMessageLib130] as SingletonDeploymentJSON[];
+
+export {
+  _ACCESSOR_DEPLOYMENTS,
```
nlordell, line 80: Why `_` prefixes?

src/types.ts
```diff
@@ -30,10 +29,10 @@ export interface SingletonDeploymentJSON {
   // 1.3.0: canonical, eip155, zksync
   // 1.4.1: canonical, zksync
   // Ex: deployments: { "canonical": { "codeHash": "0x1234", "address": "0x5678"}}
-  deployments: Record<string, Record<'address' | 'codeHash', string>>;
+  deployments: AtLeastOne<Record<AddressType, Record<'address' | 'codeHash', string>>>;
```
nlordell, line 32: nit:
```suggestion
  deployments: AtLeastOne<Record<AddressType, { address: string, codeHash: string }>>;
```

src/types.ts
```diff
@@ -57,8 +56,14 @@ export interface SingletonDeployment {
...
   // There could be multiple deployment types: canonical, eip155, zksync
+  // Possible addresses per version:
+  // 1.0.0: canonical
+  // 1.1.1: canonical
+  // 1.2.0: canonical
+  // 1.3.0: canonical, eip155, zksync
+  // 1.4.1: canonical, zksync
   // Ex: deployments: { "canonical": { "codeHash": "0x1234", "address": "0x5678"}}
-  deployments: Record<string, Record<'address' | 'codeHash', string>>;
+  deployments: AtLeastOne<Record<AddressType, Record<'address' | 'codeHash', string>>>;
```
nlordell, line 66:
```suggestion
  deployments: AtLeastOne<Record<AddressType, { address: string, codeHash: string }>>;
```

src/types.ts
```diff
@@ -72,10 +77,10 @@ export interface SingletonDeploymentV2 {
   released: boolean;
   contractName: string;
   version: string;
-  deployments: Record<AddressType, Record<'address' | 'codeHash', string>>;
+  deployments: AtLeastOne<Record<AddressType, Record<'address' | 'codeHash', string>>>;
```
nlordell, line 80:
```suggestion
  deployments: AtLeastOne<Record<AddressType, { address: string, codeHash: string }>>;
```

nlordell (approved):
LGTM, one suggestion for types (I find the object notation easier to understand than the `Record` type).

src/deployments.ts
mmv08: > // The arrays are prefixed with an underscore because they are not meant to be imported directly.

katspaugh (approved):

mmv08:
I have just realised that I'm an idiot and I forgot to add documentation 🤦

PR #698, opened 2024-07-16
Refactor GitHub Review Script

nlordell:
This PR splits the GitHub review script into two pieces:
1. The old review script, that takes a PR and verifies it meets all the requirements for deployment PRs
2. A new Node.JS script that verifies the actual deployment and code, but does not check all the PR rules.

The reason for this split is to accomodate PRs like #695, where the RPC url is not included in PR description, and more than one version is added at a time. It essentially adds a "manual" network verification scheme that can be used in special cases the PR above.

Additionally, it removes the requiment on `cast`. This is particularly important for automated checks, as we do not want to depend on nightly tools that don't have releases with somewhat guaranteed behaviour.

Lastly, this PR does a couple minor fixes to the review script, in particular it causes it to fail if more than one version is changed in the PR, and doesn't print usage information on general errors.

### DIFF
--- a/bin/github-review.sh
+++ b/bin/github-review.sh
@@ -26,10 +26,6 @@ if ! command -v gh &> /dev/null; then
     echo "ERROR: Please install the 'gh' GitHub CLI" 1>&2
     exit 1
 fi
-if ! command -v cast &> /dev/null; then
-    echo "ERROR: Please install the 'cast' tool included in the Foundry toolset" 1>&2
-    exit 1
-fi
 
 if [[ "$#" -ne 1 ]]; then
     echo "ERROR: Invalid number of arguments" 1>&2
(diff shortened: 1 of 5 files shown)


bin/github-review.sh
```diff
@@ -26,10 +26,6 @@ if ! command -v gh &> /dev/null; then
     echo "ERROR: Please install the 'gh' GitHub CLI" 1>&2
     exit 1
 fi
-if ! command -v cast &> /dev/null; then
-    echo "ERROR: Please install the 'cast' tool included in the Foundry toolset" 1>&2
-    exit 1
-fi
```
nlordell, line 32: `cast` is no longer used.

bin/github-review.sh
```diff
@@ -38,44 +34,27 @@ if [[ "$#" -ne 1 ]]; then
...
+if [[ -z $chainid ]]; then
     echo "ERROR: Chain ID not specified as per the PR Template" 1>&2
-    usage
-    exit 1
-fi
-if ! [[ $prChainID =~ ^[0-9]+$ ]]; then
-    echo "ERROR: $prChainID is not a valid Chain ID number" 1>&2
-    usage
-    exit 1
-fi
```
nlordell, line 55: This check is tautological, since the `chainid` value is read to match this pattern.

bin/github-review.sh
```diff
@@ -38,44 +34,27 @@ if [[ "$#" -ne 1 ]]; then
...
-chainlistURL="https://chainlist.org/chain/$prChainID"
-if ! curl --fail -s -o /dev/null "$chainlistURL"; then
```
nlordell, line 57: chainlist conformance gets checked in `verify` code.

bin/github-review.sh
```diff
@@ -38,44 +34,27 @@ if [[ "$#" -ne 1 ]]; then
...
-    usage
     exit 1
 fi
-chainid="$(cast chain-id --rpc-url "$rpc")"
-if [[ $chainid != $prChainID ]]; then
-    echo "ERROR: RPC $rpc doesn't match chain ID $prChainID" 1>&2
-    usage
+version="$(gh pr diff $pr --name-only | sed -nE 's|^src/assets/v([0-9\.]*)/.*$|\1|p' | sort -u)"
+if [[ "$(echo "$version" | wc -w)" -ne 1 ]]; then
+    echo "ERROR: Exactly one version must be added per PR" 1>&2
```
nlordell, line 52: Handle multiple versions per PR more gracefully.

scripts/verify.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  return {
+    ...(values as Options),
+    verbose: values.verbose === true,
+  };
+}
+
+async function main() {
+  const options = parseOptions();
+  const debug = (...msg: unknown[]) => {
```
remedcu, line 39: Doubt: Can the type be `string[]`?

remedcu (approved):
Small nit, rest LGTM 👍🏾

scripts/verify.ts
mmv08: It can also take other JavaScript types: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#Data_types

mmv08 (approved):
image

scripts/verify.ts
```diff
@@ -0,0 +1,93 @@
...
+    const asset = path.join(assets, file);
+    const json: SingletonDeploymentJSON = JSON.parse(await fs.readFile(asset, 'utf-8'));
+    debug(`${json.contractName} deployments:`);
+
+    const deployments = json.networkAddresses[options.chainId];
+    if (deployments === undefined) {
+      throw new Error(`missing ${json.contractName} deployment`);
+    }
+
+    for (const deployment of [deployments].flat()) {
```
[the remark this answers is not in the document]
nlordell: Makes sense, will change.

scripts/verify.ts
nlordell: Not in this case. In particular, this is the type of `console.log` style functions ([source](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/47c5b50ec30234ae2ee0c9ff6cdb658604941dbe/types/node/console.d.ts#L227))

```ts
log(message?: any, ...optionalParams: any[]): void;
```

Meaning that it accepts any `message` or optional parameters. We further narrow the type to `unknown` to prevent misuse of it in our `debug` function implementation.

Why? So we can do stuff like:
```ts
debug({ someObject: "with fields" });
```

PR #728, opened 2024-08-01
Add Coston2 v1.3

jO-Osko:
- Chain_ID: 114
- RPC_URL: https://coston2-api.flare.network/ext/bc/C/rpc


remedcu:
The chain ID order is wrong, please rectify that.

remedcu (approved):

PR #730, opened 2024-08-04
filecoin calibration deployment addresses

xBalbinus:
## Add new chain

- Chain_ID: 314159

- RPC_URL: https://api.calibration.node.glif.io/rpc/v1

Relevant information:


remedcu:
At the moment, we only support 1 Chain ID per PR. Could you make another PR for the different chain ID?

xBalbinus:
Sure thing

PR #751, opened 2024-08-20
Deployment diff review from bash to TS

### LINKED ISSUE #742: Deployment script verification for the same chain
This issue captures the work required to make changes to the verification script to include checks for additional deployments to the same chain.

The current script doesn't recognize these cases correctly. Example: #740

### LINKED ISSUE #744: Port Asset File Diff Verification to TypeScript
Currently, the logic to check for new PRs that the proposed changes to the asset files are OK is done in Bash here:

https://github.com/safe-global/safe-deployments/blob/0fa46dbe3f70c0815a13ead5ce1007b177d474df/bin/github-review.sh#L70-L133

The logic is becoming complex enough, that we should consider porting this to TypeScript to make the code more maintainable and easier to understand. We can also leverage additional tools such as the one @akshay-ap found (`json-diff`) as well as even right unit tests for some of the harder pieces of logic (not required for the issue, but if it helps during development then it is suggested).

remedcu:
This PR fixes #742 and #744 by porting file change checks to TS. Currently, two different checks are performed:
- When a chain with the largest chain ID is added
- When a new deployment is added to an already present chain ID

This script only checks based on the patch and the number of additions and deletions. The rest of the checks are divided between the bash and `review:verify-deployment` scripts as before.

A new package is used for working with diff called [`parse-diff`](https://www.npmjs.com/package/parse-diff).


bin/github-review.sh
```diff
@@ -69,68 +69,8 @@ fi
...
-                fi
-                # Reset the pattern_correct flag for the next set of lines
-                pattern_correct=true
-                ;;
-        esac
-        # Increment the counter
-        ((counter++))
-    done
-fi
+diffPatchSeparated=$(gh pr diff $pr | grep -E '^[+-] ' | base64) # Sending multiple lines as a string through base64 encoding
```
mmv08, line 72: Why encode it as base64?

mmv08 (approved):

bin/github-review.sh
mmv08: Ah, is this because environment variables are expected to be single line?

remedcu: Yes

nlordell: I think you can quickly run into character limits with this. Personally, I think the "correct" approach would be to either:

- use the `ChildProcess` API in the script to compute the diff
- use files. You can even make the script read the diff over STDIN to the node process. An easy approach would be to either:
```
gh pr diff $pr > .temp-github-review.diff
npm run verifyChanges -s -- --fileLineChange "$fileLineChangeJSON" --diffPatch .temp-github-review.diff  --verbose
```

I'm also not a fan of processing the input before passing it to the script instead of just having the script filter lines its interested in (i.e. do the `grep -E '^[+- ] '` in JavaScript). My rationale is that the script then accepts a standardized format (diff format) instead of a bespoke one.

Same comment for the JSON input.

package.json
```diff
@@ -19,6 +19,7 @@
     "lint:fix:ts": "npm run lint:ts -- --fix",
     "lint:fix:json": "prettier -w src/assets/*/*.json",
     "verify": "ts-node scripts/verify.ts",
+    "verifyChanges": "ts-node scripts/verifyChange.ts",
```
nlordell, line 22: These two are confusing, can we rename the first one so we have clearer "what each script verifies". As a suggestion:

```suggestion
    "review:verify-deployment": "ts-node scripts/review/verifyDeployment.ts",
    "review:diff": "ts-node scripts/review/diff.ts",
```

nlordell: (the term verify means more "verify on-chain" to me).

scripts/verifyChange.ts
```diff
@@ -0,0 +1,120 @@
...
+    } else if (additions === 1 && deletions === 1) {
+      additionalDeploymentToSameChainID = true;
+    } else if (additions !== 1 || deletions !== 0) {
+      throw new Error(`ERROR: ${path} has invalid changes`);
+    }
+  }
+
+  if (highestChainID) {
+    debug('Highest chain ID deployment');
+    let i = 0;
```
nlordell, line 59: Nit: now that we have a proper language, we can use a state enum here:

```
enum State {
  Removal,
  AddingComma,
  NewChain
}
```

scripts/verifyChange.ts
```diff
@@ -0,0 +1,120 @@
+import util from 'node:util';
+
+type Options = {
+  fileLineChange: string;
+  diffPatchSeparated: string;
```
nlordell, line 5: On the one hand, I'm not a fan of having two separate inputs here with duplicated data. It means that technically, the script can be run with inputs that don't line up (i.e. you can pass in a file line change JSON that doesn't align with the patch that was provided and the script should handle these inconsistencies).

You can probably write a "poor man's diff parser", or just use something like https://www.npmjs.com/package/parse-diff to give a structured parsing of the diff.

I think the main advantage of going to JS/TS is that we have a lot more tools available and so we don't need to do similar hacks/short-cuts as we did in bash.

bin/github-review.sh
remedcu: Nice idea, updated 👍🏾

package.json
remedcu: Done 👍🏾

scripts/verifyChange.ts
remedcu: Updated 👍🏾

scripts/verifyChange.ts
remedcu: Using `parse-diff` now 👍🏾

bin/github-review.sh
```diff
@@ -68,9 +68,9 @@ if [[ -n "$(gh pr diff $pr --name-only | grep -v -e 'src/assets/v'$version'/.*\.
 fi
 
 echo "Checking changes to assets files"
-fileLineChangeJSON=$(gh pr view $pr --json files) # This line fetches the JSON output of the files changed in the PR
-diffPatchSeparated=$(gh pr diff $pr | grep -E '^[+-] ' | base64) # Sending multiple lines as a string through base64 encoding
-npm run verifyChanges -s -- --fileLineChange "$fileLineChangeJSON" --diffPatchSeparated "$diffPatchSeparated"  --verbose
+gh pr diff $pr > .temp-github-review.diff # This line fetches the diff output of the PR
+npm run review:diff -s -- --diffPatchFileName .temp-github-review.diff --verbose
+rm .temp-github-review.diff
```
nlordell, line 73: If you want to cleanup the file more reliably after running the script, then you should use `trap`. Personally, I would say "screw it" and just add it to the project's .gitignore.

scripts/review/diff.ts
```diff
@@ -0,0 +1,143 @@
...
+
+enum HighestChangeType {
+  Initial,
+  Removal,
+  Addition,
+}
+
+function parseOptions(): Options {
+  const options = {
+    diffPatchFileName: { type: 'string' },
```
nlordell, line 18: uber-naming-nit: feels overly verbose
```suggestion
    patch: { type: 'string' },
```

PS: Leave as is if you disagree.

scripts/review/diff.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  // Read the content of the files
+  const diffPatchContent = await fs.readFile(options.diffPatchFileName, 'utf-8');
+  const diffPatch = parseDiff(diffPatchContent);
+
+  let highestChainID = false;
+  let additionalDeploymentToSameChainID = false;
+
+  for (const file of diffPatch) {
+    const { to, from, additions, deletions } = file;
```
nlordell, line 50: ubernit:

```suggestion
  for (const { to, from, additions, deletions } of diffPatch) {
```

scripts/review/diff.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    // Check to see if the file name was changed.
+    if (to !== from) {
+      throw new Error(`ERROR: ${to} and ${from} are different`);
```
nlordell, line 54:
```suggestion
      throw new Error(`ERROR: file renamed to ${to} from ${from}`);
```

scripts/review/diff.ts
```diff
@@ -0,0 +1,143 @@
...
+    }
+  }
+
+  if (highestChainID) {
+    debug('Highest chain ID deployment');
+    for (const file of diffPatch) {
+      const { chunks } = file;
+
+      for (const chunk of chunks) {
+        const { changes } = chunk;
```
nlordell, line 75:
```suggestion
    for (const { chunks } of diffPatch) {
      for (const { changes } of chunks) {
```

scripts/review/diff.ts
```diff
@@ -0,0 +1,143 @@
...
+
+        for (const change of changes) {
+          const { type, content } = change;
+
+          if (type === 'del') {
+            if (state !== HighestChangeType.Initial) {
+              throw new Error('Removal is not in correct order');
+            }
+            state = HighestChangeType.Removal;
+            previousLine = '+'.concat(content.slice(1)).concat(',');
```
nlordell, line 87: naming-nit: This is the expected next line and not the previous line right?
```suggestion
            expectedNextLine = '+'.concat(content.slice(1)).concat(',');
```

nlordell: Also, why is the `+` needed? shouldn't the diff parser already strip out the +/- prefix?

scripts/review/diff.ts
```diff
@@ -0,0 +1,143 @@
...
+    debug('Highest chain ID deployment');
+    for (const file of diffPatch) {
+      const { chunks } = file;
+
+      for (const chunk of chunks) {
+        const { changes } = chunk;
+        let state = HighestChangeType.Initial;
+        let previousLine = '';
+
+        for (const change of changes) {
```
nlordell, line 79: This feels like it shouldn't be a loop, but instead:

1. assert(changes.length is 3)
2. assert(changes.map(c => c.type) is [del, add, add])
2. assert(`${changes[0].contents},` is changes[1].contents)

scripts/review/diff.ts
```diff
@@ -0,0 +1,143 @@
...
+  if (additionalDeploymentToSameChainID) {
+    debug('Additional deployment to same chain ID');
+    let previousDeployment: string[] = [];
+    for (const file of diffPatch) {
+      const { chunks } = file;
+
+      for (const chunk of chunks) {
+        const { changes } = chunk;
+
+        for (const change of changes) {
```
nlordell, line 114: Like above, I don't think this should be a for loop.

scripts/review/diff.ts
```diff
@@ -0,0 +1,143 @@
...
+          const { type, content } = change;
+
+          if (type === 'del') {
+            // Only one deployment was present.
+            if (content.search('\\[') === -1) {
```
nlordell, line 119: Nit: instead of manual string parsing, we can:

```js
JSON.parse(`{${content}}`)
```

As it is, we don't check that the chainID didn't change:

```diff
-  "1234": "canonical",
+  "1235": ["canonical", "eip155"],
```

bin/github-review.sh
remedcu: Why not both 😅 TIL `trap`.

scripts/review/diff.ts
remedcu: Leaving it as it is 😁

scripts/review/diff.ts
remedcu: TIL 👍🏾 

Updated.

scripts/review/diff.ts
remedcu: Updated 👍🏾

scripts/review/diff.ts
remedcu: Updated 👍🏾

scripts/review/diff.ts
remedcu: The diff parser doesn't strip those out in `content`.

scripts/review/diff.ts
remedcu: Updated in the way mentioned above 👍🏾

scripts/review/diff.ts
remedcu: Done 👍🏾

scripts/review/diff.ts
remedcu: Updated, thanks! 👍🏾

scripts/review/diff.ts
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  if (highestChainID) {
+    debug('Highest chain ID deployment');
+    for (const { chunks } of diffPatch) {
+      // Filter chunks as `changes` to only include items that are additions (add) or deletions (del).
+      const changes = chunks.flatMap(({ changes }) => changes.filter(({ type }) => type === 'add' || type === 'del'));
+      assert(changes.length === 3);
+      assert(changes.map((c) => c.type).join() === 'del,add,add');
+      assert(changes[0].content.slice(1) === changes[1].content.slice(1, -1));
```
nlordell, line 71: nit: We can be a bit more precise here I think:

```suggestion
      assert(changes[0].content.replace(/^-(.*)/, "+$1,") === changes[1].content);
```

This basically means: replace leading `-` with a `+` and append a comma. This way we are really comparing exact things.

scripts/review/diff.ts
```diff
@@ -0,0 +1,113 @@
...
+    }
+  }
+
+  if (highestChainID) {
+    debug('Highest chain ID deployment');
+    for (const { chunks } of diffPatch) {
+      // Filter chunks as `changes` to only include items that are additions (add) or deletions (del).
+      const changes = chunks.flatMap(({ changes }) => changes.filter(({ type }) => type === 'add' || type === 'del'));
+      assert(changes.length === 3);
+      assert(changes.map((c) => c.type).join() === 'del,add,add');
```
nlordell, line 70: I'm not a fan of this, since it claims that `["del,add", "add"] (equals) ["del", "add,add"]`.

You can introduce a function for this:

```js
function hasChangeTypes(chunks, types) {
  return chunks.length === types.length
    && chunks.every(({ type }, i) => type === types[i]);
}
```

This can be reused below as well.

nlordell: Also, with this function, you don't need to check the length since it is implicit here.

scripts/review/diff.ts
```diff
@@ -0,0 +1,113 @@
...
+      const changes = chunks.flatMap(({ changes }) => changes.filter(({ type }) => type === 'add' || type === 'del'));
+      assert(changes.length === 2);
+      assert(changes.map((c) => c.type).join() === 'del,add');
+
+      // Read values from old and new deployments.
+      const oldDeployments = Object.values(JSON.parse(`{${changes[0].content.slice(1, -1)}}`))[0];
+      const newDeployments = Object.values(JSON.parse(`{${changes[1].content.slice(1, -1)}}`))[0];
+
+      // New deployment should always be more than one.
+      if (typeof newDeployments !== 'object') {
```
nlordell, line 88: Shouldn't this be `array`?

nlordell (commented):
Looks mostly good, left some more comments.

scripts/review/diff.ts
remedcu: Done 👍🏾

scripts/review/diff.ts
```diff
@@ -0,0 +1,113 @@
...
+  if (additionalDeploymentToSameChainID) {
+    debug('Additional deployment to same chain ID');
+    for (const { chunks } of diffPatch) {
+      const changes = chunks.flatMap(({ changes }) => changes.filter(({ type }) => type === 'add' || type === 'del'));
+      assert(changes.length === 2);
+      assert(changes.map((c) => c.type).join() === 'del,add');
+
+      // Read values from old and new deployments.
+      const oldDeployments = Object.values(JSON.parse(`{${changes[0].content.slice(1, -1)}}`))[0];
+      const newDeployments = Object.values(JSON.parse(`{${changes[1].content.slice(1, -1)}}`))[0];
```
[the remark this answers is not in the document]
remedcu: Done 👍🏾

scripts/review/diff.ts
```diff
@@ -8,7 +8,7 @@ type Options = {
   verbose: boolean;
 };
 
-type Deployment = { [key: number]: string };
+type Deployments = { [chainID: number]: string | string[] };
```
nlordell, line 11: ubernit: Typically, in JS we use camelCase with lower-case acronyms: 
```suggestion
type Deployments = { [chainId: number]: string | string[] };
```

remedcu: Done 👍🏾

scripts/review/diff.ts
```diff
@@ -0,0 +1,118 @@
...
+    if (options.verbose) {
+      console.debug(...msg);
+    }
+  };
+
+  // Read the content of the files
+  const diffPatchContent = await fs.readFile(options.diffPatchFileName, 'utf-8');
+  const diffPatch = parseDiff(diffPatchContent);
+
+  let highestChainID = false;
```
nlordell, line 47: I hadn't noticed this before - Personally, I prefer camelCase with lower-case for acronyms (so `Id`, `Http`).

This is about my suggestion :point_up:, I think consistency matters more than what we chose - so feel free to go back to `chainID` or change `ID` -> `Id` in other places as well.

remedcu: Done 👍🏾

nlordell (approved):

PR #760, opened 2024-09-02
Update Zircuit Mainnet addresses

leonimella:
## Add new chain

- Chain_ID: 48900

- RPC_URL: https://zircuit1-mainnet.p2pify.com

Relevant information:


nlordell:
We do not support removing existing deployments - you can add it by changing the value to `["canonical", "eip155"]`.

nlordell (changes requested):
We do not support removing existing deployments - you can add it by changing the value to `["canonical", "eip155"]`.

leonimella:
> We do not support removing existing deployments - you can add it by changing the value to `["canonical", "eip155"]`.

Thank you for the clarification, @nlordell. I updated the PR.

remedcu (approved):

nlordell (approved):

PR #767, opened 2024-09-09
feat: add missing deployments to 1.1.1 and 1.2.0

jmealy:
## Add new chain

- Chain_ID: 

- RPC_URL: 

Relevant information:


jmealy:
We will no longer offer multichain for older safes, closing

PR #782, opened 2024-09-16
Split workflows

akshay-ap:
The current github `release` workflow has 3 use cases:
1. Create PR to bump version
2. Create draft release
3. Publish draft release to Github if already published to NPM

Using this workflow is not intuitive and can lead to confusion as multiple actions (bumping the version, creating a draft release, and publishing the release) are combined into a single workflow.

Changes in PR:
- Split github `release` workflow into following:
  - Bump version workflow
  - Create draft release workflow
  - Publish release to github workflow

This separation increases the clarity, maintainability, and flexibility of the release process.

### DIFF
--- a/.github/workflows/bump_version.yml
+++ b/.github/workflows/bump_version.yml
@@ -1,0 +1,24 @@
+name: Bump Version
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+      - run: |
+          npm ci
+          bash bin/github-release.sh --verbose bump
+        env:
+          GH_TOKEN: ${{ github.token }}
--- a/.github/workflows/draft_release.yml
+++ b/.github/workflows/draft_release.yml
@@ -1,4 +1,4 @@
-name: Release
+name: Draft Release
 on:
   workflow_dispatch:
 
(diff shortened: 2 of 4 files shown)


bin/github-draft-release.sh
```diff
@@ -129,14 +106,4 @@ elif [[ "$current" != "$latest" ]]; then
 		fi
 		gh release create "$tag" --draft --generate-notes --target "$commit" --title "$tag" "$package"
 	fi
-else # if [[ -n "$draft" ]] && [[ "$current" == "$latest" ]]; then
```
nlordell, line 132: I think we should add a "please publish release before creating new draft".

Overwriting the draft is bad, we would lose the commit the NPM package was published with.

bin/bump-version.sh
```diff
@@ -0,0 +1,107 @@
...
+fi
+
+tag="v$current"
+commit="$(git rev-parse HEAD)"
+draft="$(gh release view "$tag" --json isDraft,targetCommitish --jq 'if(.isDraft) then .targetCommitish else "" end' 2>/dev/null || true)"
+
+log "commit:  ${commit}"
+log "draft:   ${draft:-none}"
+
+if [[ -z "$draft" ]] && [[ "$current" == "$latest" ]]; then
```
nlordell, line 78: There is so much duplicate code here...

Can't we just change the `github-release.sh` script to have verbs and then create separate GitHub actions with each verb?

For example:

```
bash bin/github-release.sh bump
```

.github/workflows/bump_version.yml
```diff
@@ -1,4 +1,4 @@
-name: Release
+name: Bump Version
 on:
   workflow_dispatch:
```
remedcu, line 3: Can we [schedule](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule) the workflow to run on a timeline (like every day) so it doesn't have to be done manually?

remedcu (approved):
Small suggestion if the GitHub token can bump and draft release on schedule.

.github/workflows/bump_version.yml
akshay-ap: If we schedule version bump, we also need to publish the release regularly because version bump will not work if a version is not published to NPM registry. I think we should have a set policy on how and when we do releases if we schedule it. I will merge as it is for now.

PR #847, opened 2024-10-28
Delay between checks

remedcu:
Sometimes when multiple deployments exist for a single chain, RPC with lower limits (free tier or lower capacity) results in a timeout.

To avoid this issue, we are adding a delay of 1 sec (which can be increased later, if found that this delay is not enough) between each file is checked.

### DIFF
--- a/scripts/review/verifyDeployment.ts
+++ b/scripts/review/verifyDeployment.ts
@@ -1,7 +1,7 @@
 import { promises as fs } from 'node:fs';
 import path from 'node:path';
 import util from 'node:util';
-import { setTimeout } from 'node:timers/promises';
+import { setTimeout as sleep } from 'node:timers/promises';
 
 import { ethers } from 'ethers';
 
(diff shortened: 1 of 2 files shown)


scripts/review/verifyDeployment.ts
```diff
@@ -84,6 +84,7 @@ async function main() {
 
       debug(`• ${deployment} deployment OK`);
     }
+    await new Promise((resolve) => setTimeout(resolve, 1000));
```
mmv08, line 87: > 87

From Node.js 15 and up you can use the [Timers Promises API](https://nodejs.org/api/timers.html#timers-promises-api). You don't have to promisify setTimeout or rely on a 3rd party library anymore.

import { setTimeout } from 'node:timers/promises';

await setTimeout(1000);

source:
https://stackoverflow.com/a/71442573/7820085

mmv08: with this you can write in a way where it reads much better, example:
```
import { setTimeout as sleep } from 'node:timers/promises';

await sleep(1000)
```

remedcu: Updated 👍🏾

mmv08 (approved):

PR #867, opened 2024-11-07
Fix Release Script Draft Creation

nlordell:
I noticed when inspecting the GitHub release script that it was no longer creating new draft releases if a previous one already exists. This is undesirable behaviour, as it does not let us accumulate changes for a draft release while the publish is pending. If we switch to running the draft GitHub action more often, then this will become an issue.

Additionally, I refactored the script as there was too much rightward-drift for my liking.

### DIFF
--- a/bin/github-release.sh
+++ b/bin/github-release.sh
@@ -39,6 +39,10 @@ while [[ $# -gt 0 ]]; do
 		dryrun="y"
 		;;
 	bump|draft|publish)
+		if [[ -n "$command" ]]; then
+			usage
+			exit 1
+		fi
 		command="$1"
 		;;
 	*)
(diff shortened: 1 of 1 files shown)


bin/github-release.sh
```diff
@@ -90,77 +90,82 @@ draft="$(gh release view "$tag" --json isDraft,targetCommitish --jq 'if(.isDraft
...
-			fi
-		else 
-			log "skipped release publishing. Either no draft release exists or the current version is not the latest NPM release."
+command_bump () {
+	log "==> Bumping Version"
+
+	if [[ -n "$draft" ]] || [[ "$current" != "$latest" ]]; then
+		log "already on unreleased version"
+		exit 0
+	fi
```
nlordell, line 99: Refactored:

```bash
if condition; then
  # ...
else
  log "..."
fi
```

to

```bash
if ! condition; then
  log "..."
  exit 0
fi

# ...
```

remedcu (approved):

PR #872, opened 2024-11-19
ADD: add Safe smart contract in Hedera Testnet

alfredolopez80:
## Add new chain

- Chain_ID: 296

- RPC_URL: https://testnet.hashio.io/api

Relevant information:
 block explorer: https://hashscan.io/testnet/dashboard


mmv08:
ERROR: chain is not registered on Chainlist

PR #1006, opened 2025-03-12
Feat: Add v1.3.0 contracts for Lens Mainnet

akshay-ap:
## Add new chain

- Chain_ID: 232

Relevant information:
NA


mmv08 (approved):
Since the network is private, approving without checking

PR #1010, opened 2025-03-13
[FEAT] list corn deployments

jierlich:
## Add new chain

- Chain_ID: 21000000


mmv08:
We limit deployments to 1 version per PR. Could you please remove and create one of the versions as a separate PR?

jierlich:
Sure. Out of curiosity, why is it limited?

PR #1018, opened 2025-03-18
ADD: Perennial Mainnet ver 1.3.0

alfredolopez80:
## Add new chain

- Chain_ID: 1424

Relevant information:
https://explorer.perennial.foundation/


nlordell:
Weird... looks like the Chainlink URL was returning a 404, but seems fine when I just checked. Will retry.

nlordell (approved):

PR #1045, opened 2025-04-17
feat:  add  bitlayer-testnet 1.4.1 contracts

307590317:
(cherry picked from commit 943d47414aafc51c1d2af4329cf2ef5ac0543948)

## Add new chain

- Chain_ID: 200810

Relevant information:
- RPC: https://testnet-rpc.bitlayer.org/
- blockexplorer: https://testnet.btrscan.com/


src/assets/v1.4.1/compatibility_fallback_handler.json
```diff
@@ -569,4 +570,4 @@
       "type": "function"
     }
   ]
-}
+}
```
nlordell, line 573: Please do not remove the newline at the end of the file.

nlordell (changes requested):

src/assets/v1.4.1/compatibility_fallback_handler.json
307590317: ok，The newline at the end of the file has been restored.

307590317:
> Looks like the change doesn't respect sorting (we make sure that the map is sorted by chain ID).
thanks, order is ok

nlordell (approved):
Thanks for fixing all the issues!

PR #1047, opened 2025-04-17
add Bitlayer mainnet 1.3.0 and 1.4.1 contracts

307590317:
## Add new chain

- Chain_ID: 200901

Relevant information:
- RPC: https://rpc.bitlayer.org/
- blockexplorer: https://www.btrscan.com/


nlordell (changes requested):
Please split this into two PRs, we only allow a single Safe version per chain per PR.

PR #1142, opened 2025-07-07
ADD: add injective testnet v1.3.0

alfredolopez80:
## Add new chain

> Template to provide information about the new chain. Only add extra information in the bottom section. Ensure that the contracts are deployed on the chain, if not deploy with [safe-contracts](https://github.com/safe-global/safe-contracts). Only **one** chain ID, **one** Safe version and **one** deployment type per PR. The RPC will be taken from [ethereum-lists/chains](https://github.com/ethereum-lists/chains). This entire paragraph should be deleted.

Please fill the following form:

Provide the Chain ID (Only 1 chain id per PR).
- Chain_ID: 1439

Relevant information:
Block Explorer: https://testnet.blockscout.injective.network/

PR #1143, opened 2025-07-08
Network Review CI Improvements

nlordell:
Some small improvements to the review CI workflow:

* Bump the Node JS version to v22 - this is the current LTS version (this also applies to other workflows)
* Make sure the diff covers _all_ files that are not assets. This is important, as it avoids changes to the workflow file, but also the `scripts/review/verifyDeployment.js` file
* Split out the review verification and dependency installation to separate steps; for `npm ci` this has a nice security benefit that the `GH_TOKEN` is not available when installing packages. It also makes it easier to diagnose exactly what step fails in CI - (see #1142, where the CI output isn't clear if `npm ci` is exiting with code 22, or the review script is).

### DIFF
--- a/.github/workflows/review.yml
+++ b/.github/workflows/review.yml
@@ -16,15 +16,19 @@ jobs:
     - uses: actions/checkout@v4
     - uses: actions/setup-node@v4
       with:
-        node-version: 20
+        node-version: 22
+    - name: Verify Review Script
+      run: |
+        git fetch https://github.com/safe-global/safe-deployments.git main:safe-global-main
+        if ! git diff --exit-code safe-global-main -- . ':!src/assets'; then
+          echo "ERROR: GitHub review script may not be up-to-date" 1>&2
+          exit 1
+        fi
+    - name: Install Dependencies
+      run: |
+        npm ci
     - name: Review PR
       env:
         GH_TOKEN: ${{ github.token }}
       run: |
-        git fetch https://github.com/safe-global/safe-deployments.git main:safe-global-main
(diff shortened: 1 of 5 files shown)


.github/workflows/review.yml
```diff
@@ -16,15 +16,19 @@ jobs:
...
-      env:
-        GH_TOKEN: ${{ github.token }}
+        node-version: 22
+    - name: Verify Review Script
       run: |
         git fetch https://github.com/safe-global/safe-deployments.git main:safe-global-main
-        if ! git diff --exit-code safe-global-main -- bin/github-review.sh; then
-          echo "ERROR: GitHub review script is not up-to-date" 1>&2
+        if ! git diff --exit-code safe-global-main -- . ':!src/assets'; then
+          echo "ERROR: GitHub review script may not be up-to-date" 1>&2
```
remedcu, line 24: I would still keep this as before, as this particular task is to check the review script changes.

We already have a check for changes other than assets here: https://github.com/safe-global/safe-deployments/blob/bed58cc0f620c8fa08d1fd83483fc2e3a6b8cb7e/bin/github-review.sh#L67

funds123 (approved):

.github/workflows/review.yml
nlordell: You are correct - the reason I changed it was that there is a hidden dependency on the `git diff` check you mentioned being in the `github-review.sh` script. If we refactor it into a `.js` script, we would accidentally break the protection. Also the command remains simple enough IMO that the change doesn't make the CI workflow more complicated.

remedcu (approved):

PR #1158, opened 2025-07-17
Use DeFi Llama Chainlist

Zalens:
We require that the chain appears on https://chainlist.org/, and previously, this meant that the chain was included in the ethereum-lists/chains repository. However, since that repository is no longer actively maintained, and since then DeFi Llama has added mechanisms for "additional chains" to their DeFiLlama/chainlist repository, which is the code behind https://chainlist.org/.

This PR changes our scripts to use the Chainlist.org API for checking if a network is listed or not (instead of the ethereum-lists/chains repository).

### DIFF
--- a/bin/github-review.sh
+++ b/bin/github-review.sh
@@ -61,24 +61,11 @@ if ! chainlist=$(echo "$chainlist_response" | jq -e '.'); then
 fi
 
 # Extract RPC for the specified chain ID using jq
-# First check if the chain exists and has an RPC
-chain_exists=$(echo "$chainlist" | jq -r "map(select(.chainId == $chainid and (.rpc | length > 0))) | length > 0")
-if [[ $chain_exists != "true" ]]; then
-    echo "ERROR: RPC not found for chain ID $chainid in DefiLlama's chainlist" 1>&2
+rpc="$(echo "$chainlist" | jq --arg C "$chainid" -r '.[] | select((.chainId | tostring) == $C) | .rpc[0].url')"
+if [[ -z $rpc ]]; then
+    echo "ERROR: Chain is not listed on DefiLlama's ChainList" 1>&2
     exit 1
 fi
-
-# Then get the RPC
-if ! rpc=$(echo "$chainlist" | jq --arg C "$chainid" -r '.[] | select((.chainId | tostring) == $C) | .rpc[0].url'); then
-    echo "ERROR: Failed to parse RPC from chainlist response" 1>&2
-    exit 1
-fi
-
-if [[ -z $rpc || $rpc == "null" ]]; then
-    echo "ERROR: No valid RPC URL found for chain ID $chainid in DefiLlama's chainlist" 1>&2
-    exit 1
-fi
-
 version="$(gh pr diff $pr --name-only | sed -nE 's|^src/assets/v([0-9\.]*)/.*$|\1|p' | sort -u)"
 if [[ "$(echo "$version" | wc -w)" -ne 1 ]]; then
     echo "ERROR: Exactly one version must be added per PR" 1>&2
--- a/bin/github-review.sh
+++ b/bin/github-review.sh
@@ -69,7 +69,7 @@ if [[ $chain_exists != "true" ]]; then
 fi
 
 # Then get the RPC
-if ! rpc=$(echo "$chainlist" | jq -r ".[] | select(.chainId == $chainid) | .rpc[0].url"); then
+if ! rpc=$(echo "$chainlist" | jq --arg C "$chainid" -r '.[] | select((.chainId | tostring) == $C) | .rpc[0].url'); then
     echo "ERROR: Failed to parse RPC from chainlist response" 1>&2
     exit 1
 fi
--- a/bin/github-review.sh
+++ b/bin/github-review.sh
@@ -56,7 +56,7 @@ fi
 
 # Parse the chainlist response
 if ! chainlist=$(echo "$chainlist_response" | jq -e '.'); then
-    echo "ERROR: Invalid JSON response from DefiLlama's chainlist" 1>&2
+    echo "ERROR: RPC not found for chain ID $chainid in DefiLlama's ChainList" 1>&2
     exit 1
 fi
 
--- a/bin/github-review.sh
+++ b/bin/github-review.sh
@@ -47,10 +47,8 @@ if [[ -z $chainid ]]; then
 fi
 
 # Fetch RPC from DefiLlama's chainlist
-chainlist_response=$(curl -sfL 'https://chainlist.org/rpcs.json' || echo "CURL_ERROR:$?")
-if [[ $chainlist_response == CURL_ERROR:* ]]; then
-    error_code=${chainlist_response#CURL_ERROR:}
-    echo "ERROR: Failed to fetch chainlist from DefiLlama (curl error $error_code)" 1>&2
+if ! chainlist_response="$(curl -sfL 'https://chainlist.org/rpcs.json')"; then
+    echo "ERROR: Failed to fetch DeFiLlama ChainList" 1>&2
     exit 1
 fi
 
--- a/bin/github-review.sh
+++ b/bin/github-review.sh
@@ -56,7 +56,7 @@ fi
 
 # Parse the chainlist response
 if ! chainlist=$(echo "$chainlist_response" | jq -e '.'); then
-    echo "ERROR: RPC not found for chain ID $chainid in DefiLlama's ChainList" 1>&2
+    echo "ERROR: DefiLlama's ChainList returned invalid JSON" 1>&2
     exit 1
 fi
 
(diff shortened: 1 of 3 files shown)


bin/github-review.sh
```diff
@@ -45,12 +45,21 @@ if [[ -z $chainid ]]; then
...
+
+# Fetch RPC from DefiLlama's chainlist
+chainlist="$(curl -sfL 'https://chainlist.org/rpcs.json')"
+if [[ -z $chainlist ]]; then
+    echo "ERROR: Failed to fetch chainlist from DefiLlama" 1>&2
+    exit 1
+fi
+
+# Extract RPC for the specified chain ID using jq
+rpc="$(echo "$chainlist" | jq -r ".[] | select(.chainId == $chainid) | .rpc[0].url")"
```
akshay-ap, line 57: If this command fails, the scripts exits with an error code `22`. See https://github.com/safe-global/safe-deployments/actions/runs/16163105042/job/45618604201 for reference. 

So, script will never report `ERROR: RPC not found for chain ID $chainid in DefiLlama's chainlist` making it difficult to find root cause of failure. I suggest we fix the command in this PR.

bin/github-review.sh
```diff
@@ -45,12 +45,40 @@ if [[ -z $chainid ]]; then
...
+chainlist_response=$(curl -sfL 'https://chainlist.org/rpcs.json' || echo "CURL_ERROR:$?")
+if [[ $chainlist_response == CURL_ERROR:* ]]; then
+    error_code=${chainlist_response#CURL_ERROR:}
+    echo "ERROR: Failed to fetch chainlist from DefiLlama (curl error $error_code)" 1>&2
+    exit 1
+fi
+
+# Parse the chainlist response
+if ! chainlist=$(echo "$chainlist_response" | jq -e '.'); then
+    echo "ERROR: Invalid JSON response from DefiLlama's chainlist" 1>&2
```
nlordell, line 59: ubernit: 
```suggestion
    echo "ERROR: RPC not found for chain ID $chainid in DefiLlama's ChainList" 1>&2
```

scripts/review/verifyDeployment.ts
```diff
@@ -46,13 +46,21 @@ async function main() {
...
+  const response = await fetch('https://chainlist.org/rpcs.json');
+  if (!response.ok) {
+    debug(`fetching chain list failed with HTTP status ${response.status}`);
+    throw new Error(`Failed to fetch chainlist from DefiLlama`);
+  }
+  const chainlist = (await response.json()) as Array<{ chainId: number; rpcs: string[] }>;
+  if (!Array.isArray(chainlist)) {
+    throw new Error('Invalid response format from DefiLlama chainlist');
+  }
+  const chainExists = chainlist.some((chain) => chain.chainId === Number(options.chainId));
```
nlordell, line 59: nit: Just to be safe with weird overflow conditions, I would write this as:


```suggestion
  const chainExists = chainlist.some((chain) => `${chain.chainId}` === options.chainId);
```

scripts/review/verifyDeployment.ts
```diff
@@ -46,13 +46,21 @@ async function main() {
...
+  if (!chainExists) {
+    throw new Error(`Chain ${options.chainId} is not registered on DefiLlama's chainlist`);
```
nlordell, line 61:
```suggestion
    throw new Error(`Chain ${options.chainId} is not registered on DefiLlama's ChainList`);
```

nlordell (commented):
PR looks good already - just made some really minor suggestions.

I agree with @akshay-ap's find that should probably be fixed before merging (my other comments are just small nits that aren't important for the moment).

bin/github-review.sh
nlordell: Oops, its `//`. And I just checked, and there is no error:

- the `.[]` splat will create a stream of JSON objects
- the `select` will filter out only objects that match the condition (i.e. same chain ID)
- the `.rpc[0].url` application only applies to matched objects in the stream. If there are none (no chain ID), then it won't be applied.

This means that if the chain isn't there - the code as provided here works. You can check this with:

```
$ curl -sfL 'https://chainlist.org/rpcs.json' | jq -r ".[] | select(.chainId == 982371)"
$ echo $?
0
```

(where `982371` doesn't exist).

bin/github-review.sh
```diff
@@ -45,12 +45,40 @@ if [[ -z $chainid ]]; then
     echo "ERROR: Chain ID not specified as per the PR Template" 1>&2
     exit 1
 fi
-chainInfo="https://raw.githubusercontent.com/ethereum-lists/chains/refs/heads/master/_data/chains/eip155-$chainid.json"
-rpc="$(curl -sfL "$chainInfo" | jq -r '.rpc[0]')"
-if [[ -z $rpc ]]; then
-    echo "ERROR: RPC not fetched correctly from the ethereum-lists" 1>&2
+
+# Fetch RPC from DefiLlama's chainlist
+chainlist_response=$(curl -sfL 'https://chainlist.org/rpcs.json' || echo "CURL_ERROR:$?")
```
nlordell, line 50: Awesome!

bin/github-review.sh
```diff
@@ -45,12 +45,40 @@ if [[ -z $chainid ]]; then
...
+# Extract RPC for the specified chain ID using jq
+# First check if the chain exists and has an RPC
+chain_exists=$(echo "$chainlist" | jq -r "map(select(.chainId == $chainid and (.rpc | length > 0))) | length > 0")
+if [[ $chain_exists != "true" ]]; then
+    echo "ERROR: RPC not found for chain ID $chainid in DefiLlama's chainlist" 1>&2
     exit 1
 fi
+
+# Then get the RPC
+if ! rpc=$(echo "$chainlist" | jq -r ".[] | select(.chainId == $chainid) | .rpc[0].url"); then
```
nlordell, line 72: I don't know how much you can do with `jq` script injection. But to be safe, I would suggest:


```suggestion
if ! rpc=$(echo "$chainlist" | jq --arg C "$chainid" -r '.[] | select((.chainId | tostring) == $C) | .rpc[0].url'); then
```

bin/github-review.sh
nlordell: > the scripts exits with an error code 22.

I just checked, and this is a `curl` exit code for HTTP error when using the `-f` flag, and unrelated to the `jq` command.

bin/github-review.sh
```diff
@@ -45,12 +45,40 @@ if [[ -z $chainid ]]; then
...
+    echo "ERROR: Failed to parse RPC from chainlist response" 1>&2
+    exit 1
+fi
+
+if [[ -z $rpc || $rpc == "null" ]]; then
+    echo "ERROR: No valid RPC URL found for chain ID $chainid in DefiLlama's chainlist" 1>&2
+    exit 1
+fi
```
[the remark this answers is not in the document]
nlordell: I would go back to what you had. I like that you validate the JSON is valid, but checking it it exists before actually getting the RPC URL isn't valid.

bin/github-review.sh
```diff
@@ -45,12 +45,40 @@ if [[ -z $chainid ]]; then
...
-if [[ -z $rpc ]]; then
-    echo "ERROR: RPC not fetched correctly from the ethereum-lists" 1>&2
+
+# Fetch RPC from DefiLlama's chainlist
+chainlist_response=$(curl -sfL 'https://chainlist.org/rpcs.json' || echo "CURL_ERROR:$?")
+if [[ $chainlist_response == CURL_ERROR:* ]]; then
+    error_code=${chainlist_response#CURL_ERROR:}
+    echo "ERROR: Failed to fetch chainlist from DefiLlama (curl error $error_code)" 1>&2
+    exit 1
+fi
```
nlordell, line 55: nit: you can use the same pattern as below, the exit code from `-f` is not very informative


```suggestion
if ! chainlist_response="$(curl -sfL 'https://chainlist.org/rpcs.json')"; then
    echo "ERROR: Failed to fetch DeFiLlama ChainList" 1>&2
    exit 1
fi
```

bin/github-review.sh
```diff
@@ -45,12 +45,40 @@ if [[ -z $chainid ]]; then
...
+
+# Parse the chainlist response
+if ! chainlist=$(echo "$chainlist_response" | jq -e '.'); then
+    echo "ERROR: RPC not found for chain ID $chainid in DefiLlama's ChainList" 1>&2
```
nlordell, line 59:
```suggestion
    echo "ERROR: DefiLlama's ChainList returned invalid JSON" 1>&2
```

bin/github-review.sh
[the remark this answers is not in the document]
nlordell: Something like this: https://github.com/safe-global/safe-deployments/pull/1158/files#r2213586588

nlordell (approved):
Nice!

PR #1161, opened 2025-07-21
Add 1.3.0 canonical monad mainnet

Zalens:
- Chain_ID: 143


remedcu:
Could you only include the change to 1.3.0?

PR #1174, opened 2025-07-30
[New Chain]: Add Apex Fusion Mainnet v1.3.0

alfredolopez80:
## Add new chain

> Template to provide information about the new chain. Only add extra information in the bottom section. Ensure that the contracts are deployed on the chain, if not deploy with [safe-contracts](https://github.com/safe-global/safe-contracts). Only **one** chain ID, **one** Safe version and **one** deployment type per PR. The RPC will be taken from [DefiLlama's ChainList](https://chainlist.org/). This entire paragraph should be deleted.

- Chain_ID: 9069

Relevant information:
not have block explorer available at the moment

### DIFF
--- a/src/assets/v1.3.0/compatibility_fallback_handler.json
+++ b/src/assets/v1.3.0/compatibility_fallback_handler.json
@@ -227,6 +227,7 @@
     "8822": "eip155",
     "9000": ["canonical", "eip155"],
     "9001": ["canonical", "eip155"],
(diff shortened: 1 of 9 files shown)


alfredolopez80:
Guys, is possible to check with this RPC: https://partner-rpc-1.nexus.route3.dev/

nlordell (approved):

PR #1191, opened 2025-08-14
feat(polynomial-mainnet-1.3.0): added polynomial-mainnet 1.3.0

ivanpochta:
## Add new chain
https://chainlist.org/chain/8008

- Chain_ID: 8008

Relevant information:

### DIFF
--- a/src/assets/v1.3.0/compatibility_fallback_handler.json
+++ b/src/assets/v1.3.0/compatibility_fallback_handler.json
@@ -17,7 +17,6 @@
     }
   },
   "networkAddresses": {
-    "8008": ["canonical", "eip155"],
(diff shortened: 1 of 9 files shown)


src/assets/v1.3.0/compatibility_fallback_handler.json
```diff
@@ -17,6 +17,7 @@
     }
   },
   "networkAddresses": {
+    "8008": ["canonical", "eip155"],
```
nlordell, line 20: Can you make it so the chain IDs are ordered?

I fixed the test so that it properly reports the error in #1194

nlordell (changes requested):
See comment above.

nlordell (approved):

PR #1195, opened 2025-08-15
Sorting deployments fix

remedcu:
## TLDR
While reviewing #1191, I found out that the tests passed, despite the chain ID not being placed in the correct position. On further checking, it was found that because of the keys splitting with `,` (comma) for nested structure (Ex: `["canonical", "eip155"]`), it interfered with its ability to decode the chain IDs.

## LLM Description
This pull request refactors the logic used to extract and verify the order of chain IDs from the `networkAddresses` section in the asset JSON files. The new approach improves reliability by directly searching for chain ID keys using a regular expression, ensuring that the original order is preserved and that errors are thrown for missing data.

Test reliability improvements:

* Refactored the extraction of chain IDs from the `networkAddresses` section in the asset JSON to use a regular expression, improving accuracy and preserving the original order.
* Added error handling to throw descriptive errors if `networkAddresses` or chain ID keys are missing from the asset JSON.

### DIFF
--- a/src/__tests__/assets.test.ts
+++ b/src/__tests__/assets.test.ts
@@ -36,12 +36,23 @@ describe('assets/', () => {
               // We manually parse the JSON here, since ECMA `JSON.parse` will
               // always order fields with numeric keys.
               const json = await readAsset(version, file);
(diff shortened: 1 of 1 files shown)


nlordell:
Duplicate of #1194 - nice we both fixed it!

PR #1197, opened 2025-08-19
add 1.3.0,1.4.0,1.5.0 contracts for paix testnet.

liuppx:
## Add new chain

RPC_URL: https://devnet.ppaix.com

- Chain_ID: 32380

Relevant information:

block explorer: https://blockscout.ppaix.com
safe-singleton-factory issue: https://github.com/safe-global/safe-singleton-factory/issues/1146


remedcu:
> Only one chain ID, one Safe version and one deployment type per PR.

PR #1332, opened 2025-11-20
Add Safe version using github actions and script

Fbartoli:


.github/workflows/update-registry.yml
```diff
@@ -0,0 +1,82 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22
```
nlordell: Unrelated to this PR (so nothing to change here), but just wanted to make a mental note that the LTS version of Node is now 24, so we should probably change the CI to that at some point.

For now, leaving it at 22 to be consistent with the other jobs is preferred IMO.

.github/workflows/update-registry.yml
```diff
@@ -0,0 +1,82 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Safe version with deployment type'
+        required: true
+        type: choice
+        options:
+          - v1.3.0-canonical
+          - v1.3.0-eip155
```
nlordell: Future feature request: might make sense to add a `v1.3.0` that does both `eip155` and `canonical` at the same time.

scripts/update-registry.ts
```diff
@@ -0,0 +1,276 @@
...
+
+    // Contract needs to be updated
+    contractsToProcess.push({ file, path: filePath, json });
+  }
+
+  // Report pre-check results
+  if (missingDeploymentType.length > 0) {
+    console.warn(
+      `⚠️  ${missingDeploymentType.length} contract(s) don't support deployment type "${deploymentType}": ${missingDeploymentType.join(', ')}`
+    );
```
nlordell: I think you should error out here instead of continuing with a partial application. Our CI doesn't allow these partial states anyway (where we add a deployment to some files but not others).

scripts/update-registry.ts
```diff
@@ -0,0 +1,276 @@
...
+  if (alreadySupported.length === jsonFiles.length - missingDeploymentType.length) {
+    // All contracts that support this deployment type already have the chain ID
+    console.log(`\n✅ Chain ID ${options.chainId} with deployment type "${deploymentType}" is already supported for all contracts:`);
+    alreadySupported.forEach((name) => console.log(`   - ${name}`));
+    console.log(`\nNo changes needed.`);
+    return;
+  }
+
+  if (alreadySupported.length > 0) {
+    console.log(`\nℹ️  ${alreadySupported.length} contract(s) already support chain ID ${options.chainId} with "${deploymentType}":`);
```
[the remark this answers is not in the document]
nlordell: Feel free to ignore this suggestion if the newline is important for the output to look nice

nlordell (changes requested):
This is awesome!

I left a few code comments and some change requests, but this is a very welcome improvement to how we handle deployments! It would be great if we can open this up to make adding deployments for external contributors also easier (but outside the scope of this PR).

scripts/update-registry.ts
```diff
@@ -0,0 +1,272 @@
...
+}
+
+/**
+ * Sorts deployment types array according to official policy:
+ * 1. If there already is a deployment for the chain, keep it first (to avoid breaking changes)
+ * 2. If deploying both variants, prefer eip155
+ */
+function sortDeploymentTypes(types: AddressType[], existingValue?: AddressType): AddressType[] {
+  // Policy #1: If there's an existing deployment, preserve it as first
+  if (existingValue !== undefined && types.includes(existingValue)) {
```
nlordell: This function is only ever called with this path - I don't think you need it at all.

nlordell (approved):
One more suggestion to remove the `sortDeploymentTypes` function, as it isn't really used.

Other than that LGTM.

PR #1340, opened 2025-12-01
Fix script when called from GH UI.

Fbartoli:
It wasn't parsing the --version flag correctly.


.github/workflows/update-registry.yml
nlordell: Should we instead use
```suggestion
          npm run update-registry:full -- \
```

Fbartoli: I removed the :full  because the parameters are passed to the lint command instead of the update script
I also removed it from the JSON. AI slop coding.

nlordell (approved):

PR #1455, opened 2026-04-05
Add Ethereum Classic (61) to v1.4.1 registry

diega:
## Add new chain

- Chain_ID: 61

Relevant information:

- Network name: Ethereum Classic
- Explorer: https://etc.blockscout.com
- Safe version: v1.4.1
- Deployment type: canonical
- All 12 contracts verified on Blockscout (exact match)

Note: a companion PR for Mordor testnet (chain 63) is also open at #1456. Both add a line after "56" so the second to merge will need a trivial rebase.

### DIFF
--- a/src/assets/v1.4.1/compatibility_fallback_handler.json
+++ b/src/assets/v1.4.1/compatibility_fallback_handler.json
@@ -30,6 +30,7 @@
     "50": "canonical",
     "51": "canonical",
     "56": "canonical",
+    "61": "canonical",
     "71": "canonical",
     "81": "canonical",
     "88": "canonical",
--- a/src/assets/v1.4.1/create_call.json
+++ b/src/assets/v1.4.1/create_call.json
@@ -30,6 +30,7 @@
     "50": "canonical",
     "51": "canonical",
     "56": "canonical",
+    "61": "canonical",
     "71": "canonical",
     "81": "canonical",
     "88": "canonical",
--- a/src/assets/v1.4.1/multi_send.json
+++ b/src/assets/v1.4.1/multi_send.json
@@ -30,6 +30,7 @@
     "50": "canonical",
     "51": "canonical",
     "56": "canonical",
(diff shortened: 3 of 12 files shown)


Zalens (approved):
All 12 v1.4.1 canonical contracts verified on-chain for chain 61. Diff is clean — single chain ID, correct numeric sorting, PR template complete with explorer URL.

EPLJVKHPSP:
Hi @diega @Zalens - congarts on long-awaited update! Is it possible to recreate Safe ver. 1.4.1 (that was created of Mainnet first) on Ethereum Classic now? If yes, what's the easiest way?

diega:
Hi @EPLJVKHPSP, thanks!

Yes, it's possible. With the v1.4.1 contracts now deployed at canonical addresses on ETC mainnet, any Safe originally created on Ethereum mainnet with v1.4.1 can be recreated at the same address on ETC.

The easiest approach is to call `createProxyWithNonce` directly on ETC's ProxyFactory (`0x4e1DCf7AD4e460CfD30791CCC4F9c8a4f820ec67`) with the same `singleton`, `initializer`, and `saltNonce` as the original Ethereum mainnet deployment. `CREATE2` produces the same address on any chain when all inputs match.

The three parameters can be recovered by decoding the input data of the original Safe's creation transaction on Ethereum mainnet (Etherscan has a "Decode Input Data" button for that). The `singleton` should be `0x41675C099F32341bf84BFc5382aF534df5C7461a` (v1.4.1 Safe) or `0x29fcB43b46531BcA003ddC8FCB67FFE91900C762` (v1.4.1 SafeL2).

Any Safe Wallet UI that consumes `@safe-global/safe-deployments@1.37.55` or later will also pick up ETC support in its next release cycle, which should enable the standard multi-chain flows via UI.

EPLJVKHPSP:
thank you for answer! @diega

PR #1528, opened 2026-04-30
Migrate to pnpm

TimDaub:
TLDR: Migrate from npm to pnpm according to our TS/JS supply chain hardening RFC

## Method

- I mostly followed the RFC: https://www.notion.so/safe-global/RFC-TS-JS-Supply-Chain-mitigation-strategy-33c8180fe5738094bd98c6433a5093c6
- For any changes where I couldn't find instructions in the RFC I looked up authoritative sources to confirm what I was doing
- I was careful

## My sources for non-RFC-defined changes
- In `bin/github-review.sh` I had to remove the `-s` flag because [pnpm doesn't have an option to silence the output](https://github.com/pnpm/pnpm/issues/4879)
- For the `--` arguments passthrough, I found that [pnpm directly passes arguments through when they're behind the script name](https://pnpm.io/cli/run#options). I hence removed the `--` separator in all cases
- I had to replace a couple of npm subprogram usages (e.g. `npm view`, `npm pack`, etc.) and I found it hard to understand if their counter parts in pnpm would be backward compatible. I hence tried to find sources that would confirm to me that their replacement was valid:
  - In pnpm@10.x, `npm view` and `npm version` [call through to npm](https://github.com/pnpm/pnpm/blob/v10.32.1/pnpm/src/pnpm.ts#L8-L49), so they are the same
  - `pnpm pack`, from what I could tell, is a more reproducible alternative to `npm pack`. It guarantees the integrity of the tarball
  - The equivalent of `npm ci` is [`pnpm install --frozen-lockfile`](https://pnpm.io/cli/install#--frozen-lockfile)

## Manual installation of `rimraf`

An explanation of why I had to manually install `rimraf` is found in its commit message: 48b86535a6bd14b347de51172190d009b1f359ab

## Dependabot groups

I didn't know the preference of groups in the dependabot file. Hence I just copied the groups from safe-modules-deployments. Please let me know if that's OK.

## Hash-pinning the GitHub Actions

I manually went to the respective GitHub Action repository, I searched for the latest minor/patch version of the respective action, then manually resolved the commit hash to pin it in the yaml file.

- https://github.com/actions/checkout/releases/tag/v4.3.1 
  - => https://github.com/actions/checkout/commit/34e114876b0b11c390a56381ad16ebd13914f8d5
- https://github.com/actions/setup-node/releases/tag/v4.4.0 
  - => https://github.com/actions/setup-node/commit/49933ea5288caeca8642d1e84afbd3f7d6820020
- https://github.com/peter-evans/create-pull-request/releases/tag/v5.0.3
  - => https://github.com/peter-evans/create-pull-request/commit/4e1beaa7521e8b457b572c090b25bd3db56bf1c5
- https://github.com/pnpm/action-setup/releases/tag/v6.0.3
  - => https://github.com/pnpm/action-setup/commit/903f9c1a6ebcba6cf41d87230be49611ac97822e

## `onlyBuiltDependencies`

I didn't look a lot into this. No built dependencies are in the list because I didn't find any that should be in there. I pnpm installed the dependencies, ran the tests, and they seemed to pass which makes me conclude that no building of dependencies is necessary.

## Testing

I didn't test any of my changes regarding the GitHub Actions. If I should test them, let's please discuss how to do this well.

## Checklist included in Linear ticket description

- [x] package-lock.json is removed and replaced with pnpm-lock.yaml
- [x] package.json includes a packageManager field with the pnpm Corepack SHA512 hash
- [x] package.json includes engines pinning Node ≥ 22 and pnpm ≥ 10.16
- [x] pnpm-workspace.yaml exists with minimumReleaseAge: 10080, blockExoticSubdeps: true, and an explicit onlyBuiltDependencies allowlist
- [x] All npm scripts that use -- passthrough have been audited and fixed if needed
- [x] CI passes using pnpm install --frozen-lockfile
- [x] Every uses: reference in all workflow files is pinned to a full commit SHA
- [x] .github/dependabot.yml is present with cooldown: default-days: 7 and the github-actions ecosystem entry
- [x] All Dependabot and CI checks pass after the migration

Fixes PLA-1378


README.md
```diff
@@ -31,8 +31,9 @@ The addresses on the different networks and the abi files are available for each
 
 ## Install
 
-- npm - `npm i @safe-global/safe-deployments`
-- yarn - `yarn add @safe-global/safe-deployments`
+```sh
+pnpm add @safe-global/safe-deployments
```
dasanra: Here is fine we keep the three of them. These are installation instructions for those devs wanting to use the package so all of them are supported.

TimDaub: Fixed in https://github.com/safe-global/safe-deployments/pull/1528/changes/c1842bd929dc08ee7886f4ced9b3f2bb1435d39a

dasanra (approved):

